PR #3, opened 2019-02-25
WIP Add Bucket Provisioner Reconciler

copejon:


pkg/apis/objectbucket.io/v1alpha1/objectbucket_types.go
```diff
@@ -22,9 +21,12 @@ type ObjectBucketSpec struct {
 	SecurePort int `json:"securePort"`
 	// SSL true if the connection is secured with SSL, false if it is not.
 	SSL bool `json:"ssl"`
-
+	// BucketACL
```
jeffvance: If `SecurePort` has a value does that mean accessing the bucket should be via this port? And, if true, then what is the purpose of the SSL bool?

pkg/apis/objectbucket.io/v1alpha1/objectbucket_types.go
```diff
@@ -37,7 +39,9 @@ const (
 
 // ObjectBucketStatus defines the observed state of ObjectBucket
 type ObjectBucketStatus struct {
-	Controller *v1.ObjectReference
+	Controller *v1.LocalObjectReference
```
jeffvance: is including the controller in status fairly typical?

pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
```diff
@@ -1,13 +1,37 @@
...
+	// StorageClass names the StorageClass object representing the desired provisioner and parameters
 	StorageClass string
+	// BucketName (not recommended) the name of the bucket.  Caution!
+	// In-store bucket names may collide across namespaces.  If you define
+	// the name yourself, try to make it as unique as possible.
+	BucketName string `json:"bucketName,omityempty"`
+	// GenerateBucketName (recommended) a prefix for a bucket name to be
+	// followed by a hyphen and 5 random characters. Protects against
+	// in-store name collisions.
+	GeneratBucketName string `json:"generatBucketName,omitempty"`
```
jeffvance: I'm a bit behind I guess. Looks like the current design does _not_ use the OBC's namespace as part of the generated bkt name?

pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
```diff
@@ -1,13 +1,37 @@
...
+	// SSL whether connection to the bucket requires SSL authentication or not
+	SSL bool `json:"ssl"`
```
jeffvance: why is SSL needed in the OBC and OB? I would think this flag is only needed in the OB.

jeffvance: Ditto for the` CannedBucketACL`, `Versioned`, and `AdditionalConfig` fields. I would think these field also only part of the OB...

pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
```diff
@@ -1,13 +1,37 @@
...
+	//    BucketCannedACLPrivate
+	//    BucketCannedACLPublicRead
+	//    BucketCannedACLPublicReadWrite
+	//    BucketCannedACLAuthenticatedRead
+	CannedBucketACL s3.BucketCannedACL `json:"cannedBucketAcl"`
+	// Versioned determines if versioning is enabled
+	Versioned bool `json:"versioned"`
+	// AdditionalConfig gives non-AWS S3 providers a location to set
+	// proprietary config values (tenant, namespace, etc)
+	AdditionalConfig map[string]interface{} `json:"additionalConfig"`
```
jeffvance: can the Phase string constants be defined in only 1 file?

provisioner/auth/accesskeys.go
```diff
@@ -0,0 +1,9 @@
+package auth
```
jeffvance: why not v1alpha1 package?

provisioner/claim-reconciler/reconiler.go
```diff
@@ -0,0 +1,113 @@
+package object_bucket_claim_reconciler
```
jeffvance: can we name this simply `reconciler` and let the importing code alias this pkg if necessary?

provisioner/claim-reconciler/reconiler.go
```diff
@@ -0,0 +1,113 @@
...
+	RetryTimeout  time.Duration
+	RetryBackoff  int
+}
+
+func NewObjectBucketClaimReconciler(c client.Client, name string, provisioner provisioner.Provisioner, options ReconcilerOptions) *objectBucketClaimReconciler {
+	return &objectBucketClaimReconciler{
+		client:          c,
+		provisionerName: strings.ToLower(name),
+		provisioner:     provisioner,
+		retryInterval:   options.RetryInterval,
```
jeffvance: how are defaults handled?

provisioner/claim-reconciler/reconiler.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	// TODO wrap in retry logic
+	ob, s3keys, err := r.provisioner.Provision(obc)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if s3keys.AreEmpty() {
+		msg := "s3 access key or secret key is nil"
+		err = r.provisioner.Delete(obc)
```
jeffvance: Is this block saying "Provision() created a bkt without using (or returning) the s3 key-pairs, so therefore we need to delete that physical bkt now"?
What if (down the road) reclaimPolicy is "retain"? I know this is WIP but just sharing some initial thoughts...

provisioner/claim-reconciler/reconiler.go
```diff
@@ -0,0 +1,113 @@
...
+			msg = fmt.Sprintf("%s, error deleting bucket: %v", msg, err)
+		}
+		return reconcile.Result{}, fmt.Errorf("%v", err)
+	}
+
+	// TODO wap in retry logic
+	if err = r.client.Create(context.TODO(), ob); err != nil {
+		return reconcile.Result{}, err
+	}
+
```
jeffvance: //TODO create the Secret and CM...

provisioner/claim-reconciler/reconiler.go
```diff
@@ -0,0 +1,113 @@
...
+	return reconcile.Result{}, nil
+}
+
+// A simplistic check on whether this obc is a concern for this provisioner.  Likely to need
+// a more complex check down the road
+func (r *objectBucketClaimReconciler) shouldProvision(class string) bool {
+	return class == "" && class != r.provisionerName
```
jeffvance: confused... don't you want `return class == r.provisionerName` ?

provisioner/interface.go
```diff
@@ -0,0 +1,17 @@
...
+
+// Provisioner the interface to be implemented by users of this
+// library and executed by the Reconciler
+type Provisioner interface {
+	// Provision should be implemented to handle bucket creation
+	// for the target object store
+	Provision(*v1alpha1.ObjectBucketClaim) (*v1alpha1.ObjectBucket, *auth.S3AccessKeys, error)
+	// Delete should be implemented to handle bucket deletion
+	// for the target object store
+	Delete(claim *v1alpha1.ObjectBucketClaim) error
```
jeffvance: do you think we may need to pass the reclaimPolicy as another arg in `Delete()`?  Or, add it to the OBC...

provisioner/object-bucket-reconciler/reconciler.go
```diff
@@ -13,7 +13,7 @@ type ObjectBucketReconciler struct {
 
 // TODO if we decide that OBs should have their own Reconiler, then we can work out
 //  the logic for that here.  If not, this package can be deleted.
-func (r *ObjectBucketReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+func (r ObjectBucketReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
jeffvance: removing ptr implies that `Reconcile()` does not modify its receiver in any way. Another reason to keep the `*` is if the receiver struct is somewhat large. Though passing by value may be correct, I see k8s code where all methods on a type are pointers rather than values even though receiver is not modified by some of the methods. Not sure if this consistency pays off in the long run?

provisioner/provisioner.go
```diff
@@ -0,0 +1,150 @@
...
+	if options.Threadiness < DefaultThreadiness {
+		options.Threadiness = DefaultThreadiness
+	}
+	if options.ProvisionBaseInterval < DefaultRetryBaseInterval {
+		options.ProvisionBaseInterval = DefaultRetryBaseInterval
+	}
+	if options.ProvisionRetryTimeout < DefaultRetryTimeout {
+		options.ProvisionRetryTimeout = DefaultRetryTimeout
+	}
+	if options.ProvisionRetryBackoff < DefaultRetryBackOff {
```
jeffvance: are there any upper limit checks on these passed-in numbers?

pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
copejon: > why is `SSL` needed in the OBC and OB? I would think this flag is only needed in the OB.

this comes down to what we want users to be able to request.  Should they be able to ask for an https endpoint via PVC? Or should that be through storage class?

> Ditto for the` CannedBucketACL`, `Versioned`, and `AdditionalConfig` fields. I would think these field also only need to be defined in the OB...

pretty much the same as above

provisioner/auth/accesskeys.go
copejon: That's a good question.  This isn't an API object but it is a requirement for creating API objects.  I believe the k8s convention is to co-locate some of these helper structs with the CRD apis.

pkg/apis/objectbucket.io/v1alpha1/objectbucket_types.go
copejon: The sig-storage pattern is to add an annotation to the object that the controller "claims" responsibility for.  This seems a little cleaner than mutating / analyzing the annotations map[string]string.

pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
copejon: Do you mean using them for both OBC and OB phases?  In the current case, yes as they're identical.  We need to really decide on what they phases should be.

provisioner/claim-reconciler/reconiler.go
copejon: currently, the caller sets the defaults before calling this function.  I take your point though - it's probably more appropriate that it be handled here.

provisioner/claim-reconciler/reconiler.go
copejon: RE: retainPolicy - see my comment [here](https://github.com/rook/rook/pull/2625/commits/4a94959fea87008281fafcd4cca950fb80351b66#r260813528).  From what I can tell, k8s has obsolesced retainPolicy in the external storage library.

provisioner/claim-reconciler/reconiler.go
copejon: right, see L56

provisioner/claim-reconciler/reconiler.go
copejon: no, but the arg name should be changed from `class` to `provisioner`

provisioner/interface.go
copejon: This actually needs to be changed to accept an OB, not an OBC.

provisioner/object-bucket-reconciler/reconciler.go
copejon: right, that's a typo, will add.

provisioner/provisioner.go
copejon: At this point, i'm erring on the side of caution on behalf of the caller.   I'm assuming the caller is aware of the capability of the hardware their running.

pkg/apis/objectbucket.io/v1alpha1/objectbucket_types.go
copejon: This is derived from some other examples I've seen on bucket CRD implementations.  It's true that it's redundant.

pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
copejon: No, you're correct.  Not sure what I was thinking :P

jeffvance: typo (missing `e`)

provisioner/claim-reconciler/reconiler.go
```diff
@@ -1,14 +1,19 @@
...
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"github.com/yard-turkey/lib-bucket-provisioner/provisioner/auth"
+	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"strconv"
```
jeffvance: did `gofmt` change its preference for import order? I was expecting all go packages to be at the beginning.

provisioner/claim-reconciler/reconiler.go
```diff
@@ -24,13 +29,38 @@ type objectBucketClaimReconciler struct {
 
 var _ reconcile.Reconciler = &objectBucketClaimReconciler{}
 
-type ReconcilerOptions struct {
+const (
+	defaultRetryBaseInterval = time.Second * 10
+	defaultRetryTimeout      = time.Second * 360
+	defaultRetryBackOff      = 1
+
+	finalizer = "lib-bucket-provisioner/finalizer"
```
jeffvance: I think name changed to _objectbucket.io/finalizer_

provisioner/claim-reconciler/reconiler.go
```diff
@@ -24,13 +29,38 @@ type objectBucketClaimReconciler struct {
...
+
+	// Fields Names
+	bucketName      = "S3_BUCKET_NAME"
+	bucketHost      = "S3_BUCKET_HOST"
+	bucketPort      = "S3_BUCKET_PORT"
+	bucketSSL       = "S3_BUCKET_SSL"
```
jeffvance: current proposal has added `bucketURL = "S3_BUCKET_URL`

provisioner/claim-reconciler/reconiler.go
```diff
@@ -81,33 +113,89 @@ func (r *objectBucketClaimReconciler) Reconcile(request reconcile.Request) (reco
...
+				return true, fmt.Errorf("object bucket %s already exists: %v", ob.Name, err)
+			} else {
+				return true, fmt.Errorf("object bucket %s creation failed: %v", ob.Name, err)
+			}
+		}
+		return true, nil
+	})
+
+	credSecret := newCredentailsSecret(obc, s3keys)
+	err = wait.PollImmediate(r.retryInterval, r.retryTimeout, func() (done bool, err error) {
```
jeffvance: probably want a more generic _create_this_resource func which handles retries, polling, etc for Secrets, CMs, and OBs.

provisioner/provisioner.go
```diff
@@ -19,11 +19,8 @@ import (
 )
 
 const (
-	ProvisionerNamePrefix    = "object-bucket-claims.lib-bucket-provisioner/"
-	DefaultRetryBaseInterval = time.Second * 10
-	DefaultRetryTimeout      = time.Second * 360
-	DefaultRetryBackOff      = 1
-	DefaultThreadiness       = 1
+	ProvisionerNamePrefix = "object-bucket-claims.lib-bucket-provisioner/"
```
jeffvance: Has this changed?

pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
```diff
@@ -31,7 +31,7 @@ type ObjectBucketClaimSpec struct {
 	Versioned bool `json:"versioned"`
 	// AdditionalConfig gives non-AWS S3 providers a location to set
 	// proprietary config values (tenant, namespace, etc)
-	AdditionalConfig map[string]interface{} `json:"additionalConfig"`
+	//AdditionalConfig map[string]interface{} `json:"additionalConfig"`
```
jeffvance: What about naming this field a name consistent with what's used in a storage class or config map to hold extra values, eg `Parameters` or `Data` or ?

provisioner/claim-reconciler/reconiler.go
copejon: gofmt doesn't handle imports.  When use `go imports`, we can specify how we want them grouped.  We don't have any build scripts yet however.

pkg/apis/objectbucket.io/v1alpha1/objectbucket_types.go
```diff
@@ -1,45 +1,38 @@
...
 
 import (
-	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type ReclaimPolicy string
+
+const (
```
jeffvance: I would have expected these consts to be defined in core v1.

pkg/apis/objectbucket.io/v1alpha1/objectbucket_types.go
```diff
@@ -1,45 +1,38 @@
 package v1alpha1
 
 import (
-	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type ReclaimPolicy string
```
jeffvance: is there a core.V1 type for this?

provisioner/claim-reconciler/reconiler.go
```diff
@@ -87,67 +92,64 @@ func (r *objectBucketClaimReconciler) Reconcile(request reconcile.Request) (reco
...
 	if err != nil {
-		return reconcile.Result{}, err
+		return handleErr("error translating core.PersistentVolumeReclaimPolicy %q to v1alpha1.ReclaimPolicy: %v", storageClass.ReclaimPolicy, err)
 	}
 
-	if !r.shouldProvision(claimClass) {
-		return reconcile.Result{}, fmt.Errorf("claim failed shouldProvision check")
+	options := &provisioner.BucketOptions{
+		ReclaimPolicy:     reclaimPolicy,
+		ObjectBucketName:  obc.Namespace + "-" + obc.Name,
```
jeffvance: we want to use the k8s name generator here, right?

provisioner/claim-reconciler/reconiler.go
```diff
@@ -198,3 +196,30 @@ func newBucketConfigMap(ob *v1alpha1.ObjectBucket, obc *v1alpha1.ObjectBucketCla
...
+}
+
+func extractClaimClass(obc *v1alpha1.ObjectBucketClaim) (string, error) {
+	if obc.Spec.StorageClassName == "" {
+		return "", fmt.Errorf("no class for claim %q", obc.Name)
+	}
+	return obc.Spec.StorageClassName, nil
+}
+
+func translateReclaimPolicy(rp corev1.PersistentVolumeReclaimPolicy) (v1alpha1.ReclaimPolicy, error) {
```
jeffvance: Doesn't API validation handle incorrect reclaim policy values (for both Storage Classes and OBs)?

PR #43, opened 2019-03-28
Add dumb delete

copejon:
Simple delete call when an OBC is deleted.  Does not clean up cluster resources


pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
```diff
@@ -40,6 +39,9 @@ type ObjectBucketClaimSpec struct {
 	// AdditionalConfig gives providers a location to set
 	// proprietary config values (tenant, namespace, etc)
 	AdditionalConfig map[string]string `json:"additionalConfig"`
+	// ObjectBucketName is the name of the object bucket resource.  This is the authoritative
+	// determintaion for binding.
+	ObjectBucketName string
```
BlaineEXE, line 44: Correct me if I'm wrong @tnielsen or @leseb, but the way that K8s and go structs with JSON work, if there is no json info given here, Kubernetes will never see the `ObjectBucketName` field, correct? It will treat it as though it doesn't exist I think.

thotz: @copejon @jeffvance ^^

copejon: Oh dang, yes this is bug.  To what degree it affects the controller I don't know.  I'll do some investigating.

guymguym: Really? I think it works just the json name is by default capitalized instead of `objectBucketName`...
Not sure if kubernetes is doing something different, but json will marshal it -
```
$ cat a.go
package main

import (
	"encoding/json"
	"fmt"
)

type A struct {
	Name             string `json:"xxxxxxxxxxx"`
	ObjectBucketName string
}

func main() {
	a := A{Name: "Jon", ObjectBucketName: "Cope"}
	s, _ := json.Marshal(a)
	fmt.Println(string(s))
}

$ go run a.go
{"xxxxxxxxxxx":"Jon","ObjectBucketName":"Cope"}
```

jeffvance, line 44: Looking at the [design doc](https://github.com/kube-object-storage/lib-bucket-provisioner/blob/master/doc/design/object-bucket-lib.md#obc-custom-resource-after-update-by-lib) this field was supposed to exist in the OBC but be named `objectBucketRef`.

copejon: thanks @guymguym.  So that's why it didn't produce to breaking behavior.  Wouldn't hurt to tag it for the sake of convention, but it's otherwise innocuous.

copejon: True... i think that's probably due to the docs becoming stale as the code changed.  In the end I believe we decided a ObjectReference inner struct was overkill.

pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
```diff
@@ -53,8 +55,7 @@ const (
 
 // ObjectBucketClaimStatus defines the observed state of ObjectBucketClaim
 type ObjectBucketClaimStatus struct {
-	Phase           ObjectBucketClaimStatusPhase
-	ObjectBucketRef *v1.ObjectReference
+	Phase ObjectBucketClaimStatusPhase
```
jeffvance, line 58: Oh, I see the OB ref was moved from status to spec...

pkg/provisioner/manager.go
```diff
@@ -127,7 +127,6 @@ func NewProvisioner(
 		WithEventFilter(skipUpdate).
 		Complete(claimReconciler.NewObjectBucketClaimReconciler(client, ctrl.Manager.GetScheme(), provisionerName, provisioner, claimReconciler.Options{
 			RetryInterval: options.ProvisionBaseInterval,
-			RetryBackoff:  options.ProvisionRetryBackoff,
```
jeffvance, line 130: what's the rationale for no backoffs on deletes?

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -87,66 +81,50 @@ func (r *objectBucketClaimReconciler) Reconcile(request reconcile.Request) (reco
...
 
-	// ///   ///   ///   ///   ///   ///   ///
-	// TODO    CAUTION! UNDER CONSTRUCTION!
-	// ///   ///   ///   ///   ///   ///   ///
-
-	obc, err := r.claimFromKey(request.NamespacedName)
-	if err != nil && !errors.IsNotFound(err) {
-		return handleErr("error getting claim for key: %v", err)
+	// Delete path
+	if err != nil {
```
jeffvance, line 89: Is it possible for err == nil && ob.deletionTimestamp == nil?

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -226,41 +248,84 @@ func (r *objectBucketClaimReconciler) createConfigMap(options *api.BucketOptions
...
+			return false, err
+		}
+		return true, nil
+	})
+}
+
 // shouldProvision is a simplistic check on whether this obc is a concern for this provisioner.
 // Down the road, this will perform a broader set of checks.
 func (r *objectBucketClaimReconciler) shouldProvision(obc *v1alpha1.ObjectBucketClaim) bool {
-	if obc == nil {
```
jeffvance, line 239: will obc always != nil for all calls to shouldProvision()?

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -226,41 +248,84 @@ func (r *objectBucketClaimReconciler) createConfigMap(options *api.BucketOptions
...
-	class, err := util.StorageClassForClaim(obc, r.client, r.ctx)
-	if err != nil {
-		klog.Errorf("cannot provision: %v", err)
+	if obc.Spec.ObjectBucketName != "" {
+		r.logI.Info("provisioning already completed", "ObjectBucket", obc.Spec.ObjectBucketName)
 		return false
 	}
-	if class.Provisioner != r.provisionerName {
-		r.logI.Info("claim provisioner does not match expected provisioner", "claim provisioner", class.Provisioner, "should match", r.provisionerName)
+	if obc.Spec.StorageClassName == "" {
```
jeffvance, line 278: This doesn't handle the case of obc.spec.storageClass specified but the actual SC doesn't exist? This mayb e okay if this case is logged and retried.

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -226,41 +248,84 @@ func (r *objectBucketClaimReconciler) createConfigMap(options *api.BucketOptions
...
+		r.logI.Info("OBC did not provide a storage class, cannot provision")
```
jeffvance, line 279: why remove the check for SC provisioner name matching the client's provisioner name? This is how a cluster supports multiple, different object store provisioners.

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -87,66 +81,50 @@ func (r *objectBucketClaimReconciler) Reconcile(request reconcile.Request) (reco
...
 	if err != nil {
-		klog.Error(err) // controller-runtime does not report returned errors. log them before handing them off˚
-		return reconcile.Result{}, err
+		// controller-manager does not report the errors, log them before returning
+		klog.Error(err)
 	}
-
-	return reconcile.Result{}, nil
+	// If handleReconcile() errors, the request will be re-queued.  In the distant future, we will likely want some
+	// ignorable error types in order to skip re-queuing
```
jeffvance, line 121: +1

PR #51, opened 2019-04-04
add claimRef to OB, removed bucket URL from config map, unexported in…

copejon:
…ternalClient fields


pkg/provisioner/reconciler/claim-reconciler/client.go
```diff
@@ -8,9 +8,11 @@ import (
...
-// internalClient wraps related reconciler fields for convenient passing to helpers
+// internalClient wraps related reconciler fields for convenient passing to helpers. Client is a controller-runtime
+// DelegatingClient generated by the manager in manager.go. This is for optimization of API calls where in we make API
+// resource requests against the store and writes against the API server.
 type internalClient struct {
-	Ctx    context.Context
-	Client client.Client
-	Scheme *runtime.Scheme
+	client.Client
+	context context.Context
```
jeffvance: `ctx` is more typical when referring to a go context.

copejon: true, will revert

PR #63, opened 2019-04-09
use reclaimPolicy in delete choice

jeffvance:

Tested with s3 provisioner by creating an obc referencing a SC with reclaimPolicy=Retain then deleting the obc. ALl artifacts deleted and the bucket remained, per expectations. Changed SC's reclaimPolicy=Delete and bucket was deleted.

### DIFF
--- a/pkg/provisioner/reconciler/claim-reconciler/helpers.go
+++ b/pkg/provisioner/reconciler/claim-reconciler/helpers.go
@@ -30,6 +30,7 @@ func shouldProvision(obc *v1alpha1.ObjectBucketClaim) bool {
 	return true
 }
 
+<<<<<<< HEAD
 func claimRefForKey(key client.ObjectKey, ic *internalClient) (types.UID, error) {
 	claim, err := claimForKey(key, ic)
 	if err != nil {
@@ -38,10 +39,22 @@ func claimRefForKey(key client.ObjectKey, ic *internalClient) (types.UID, error)
 	return claim.UID, nil
 }
 
-// Return true if this storage class was for a new bkt vs an existing bkt.
-// referenced storage class.
-func scForNewBkt(sc *storagev1.StorageClass) bool {
+// Return true if this storage class is for a new bucket vs an existing bucket.
+func isNewBucketByClass(sc *storagev1.StorageClass) bool {
 	return len(sc.Parameters[v1alpha1.StorageClassBucket]) == 0
+}
+
+// Return true if this OB is for a new bucket vs an existing bucket.
+func (ic *internalClient) isNewBucketByOB(ob *v1alpha1.ObjectBucket) bool {
+	// temp: get bucket name from OB's storage class
+	class, err := storageClassForOB(ob, ic)
+	if err != nil || class == nil {
+		log.Info("ERROR: unable to get storageclass", "ob", ob)
+		logD.Info("ERROR: returning false for `isNewBucketByOB`")
+		return false
+	}
+	return len(class.Parameters[v1alpha1.StorageClassBucket]) == 0
+	// return ob.Spec.DynamicProivisioned
 }
 
 func claimForKey(key client.ObjectKey, ic *internalClient) (obc *v1alpha1.ObjectBucketClaim, err error) {
@@ -118,7 +131,7 @@ func generateBucketName(prefix string) string {
 	return fmt.Sprintf("%s-%s", prefix, uuid.New())
 }
 
-func storageClassForClaim(obc *v1alpha1.ObjectBucketClaim, ic *internalClient) (*storagev1.StorageClass, error) {
+func (ic *internalClient) storageClassForClaim(obc *v1alpha1.ObjectBucketClaim) (*storagev1.StorageClass, error) {
 	logD.Info("getting storageClass for claim")
 	if obc == nil {
 		return nil, fmt.Errorf("got nil ObjectBucketClaim ptr")
--- a/pkg/provisioner/reconciler/claim-reconciler/helpers.go
+++ b/pkg/provisioner/reconciler/claim-reconciler/helpers.go
@@ -30,7 +30,6 @@ func shouldProvision(obc *v1alpha1.ObjectBucketClaim) bool {
 	return true
 }
 
-<<<<<<< HEAD
 func claimRefForKey(key client.ObjectKey, ic *internalClient) (types.UID, error) {
 	claim, err := claimForKey(key, ic)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -144,6 +143,9 @@ func (r *ObjectBucketClaimReconciler) handleProvisionClaim(key client.ObjectKey,
 		return err
 	}
 
+	// use storage class to determine if this is a new or an old bucket
+	newBkt := scForNewBkt(class)
```
copejon: bit of a nit - for readability down the line, the func name would be better spelled out.  `newBkt` makes me think it's a bucket object.  maybe something like `isProvisionedBucket` or `isNewBucket`

copejon:
> Changed SC's reclaimPolicy=Delete and bucket was deleted

This brings up a really important point.  The SC should only prescribe the OB reclaim policy at point of inception.  On deletes, the _OB's_ reclaimPolicy should be the authoritative field for policy.  Changes to the SC should not overrule policies of OBs created before the change.

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -262,10 +267,13 @@ func (r *ObjectBucketClaimReconciler) handleDeleteClaim(key client.ObjectKey) er
 	if err != nil || class == nil {
 		return fmt.Errorf("error getting storageclass from OB %q", ob.Name)
 	}
+
+	// use storage class to determine if this is a new or an old bucket
```
jeffvance: a _dynamic provisioned_ flag should be stored in the OB. This is how we should decide greenfield vs. brownfield.

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -262,10 +267,13 @@ func (r *ObjectBucketClaimReconciler) handleDeleteClaim(key client.ObjectKey) er
 	if err != nil || class == nil {
 		return fmt.Errorf("error getting storageclass from OB %q", ob.Name)
```
jeffvance: we don't need to get the SC. The OB's reclaim policy should be the authoritative source for reclaim.

pkg/provisioner/reconciler/claim-reconciler/helpers.go
```diff
@@ -38,12 +38,24 @@ func claimRefForKey(key client.ObjectKey, ic *internalClient) (types.UID, error)
...
-// Return true if this storage class was for a new bkt vs an existing bkt.
-// referenced storage class.
-func scForNewBkt(sc *storagev1.StorageClass) bool {
+// Return true if this storage class is for a new bucket vs an existing bucket.
+func isNewBucketByClass(sc *storagev1.StorageClass) bool {
 	return len(sc.Parameters[v1alpha1.StorageClassBucket]) == 0
 }
 
+// Return true if this OB is for a new bucket vs an existing bucket.
+func (ic *internalClient) isNewBucketByOB(ob *v1alpha1.ObjectBucket) bool {
```
copejon: the internalClient should be a shallow encapsulation of a client.Client coupled with a context the controller's scheme.  This should be a helper that takes a `client.Client` interface and then be passed an internalClient.

pkg/provisioner/reconciler/claim-reconciler/helpers.go
```diff
@@ -118,7 +130,7 @@ func generateBucketName(prefix string) string {
 	return fmt.Sprintf("%s-%s", prefix, uuid.New())
 }
 
-func storageClassForClaim(obc *v1alpha1.ObjectBucketClaim, ic *internalClient) (*storagev1.StorageClass, error) {
+func (ic *internalClient) storageClassForClaim(obc *v1alpha1.ObjectBucketClaim) (*storagev1.StorageClass, error) {
```
copejon: ditto above wrt help func.

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -144,6 +143,10 @@ func (r *ObjectBucketClaimReconciler) handleProvisionClaim(key client.ObjectKey,
 		return err
 	}
 
+	// use storage class to determine if this bkt was dynamically provision
+	// (greenfield) or not (brownfield)
```
copejon: nits 
- `bkt was dynamically provision` should read `bucket should dynamically provisioned`
- "greenfield" and "brownfield" aren't very descriptive of the cases we're dealing with and could be hard to interpret by other devs (assuming anyone besides us ever reads this code :P )

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -202,6 +205,10 @@ func (r *ObjectBucketClaimReconciler) handleProvisionClaim(key client.ObjectKey,
 	ob.Spec.StorageClassName = obc.Spec.StorageClassName
 	ob.Spec.ClaimRef, err = claimRefForKey(key, r.internalClient)
 	ob.SetFinalizers([]string{finalizer})
+	// allow provisioner to modify reclaimPolicy
+	if ob.Spec.ReclaimPolicy == nil {
+		ob.Spec.ReclaimPolicy = options.ReclaimPolicy
```
copejon: I'm thinking this probably shouldn't be conditional.  Rather than letting provisioners set the policy, we should enforce the one defined by the storage class.  Otherwise the SC isn't representational of the backing OBs.

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -259,15 +266,18 @@ func (r *ObjectBucketClaimReconciler) handleDeleteClaim(key client.ObjectKey) er
 		log.Error(nil, "got nil objectBucket, assuming deletion complete")
 		return nil
 	}
+	// see if obc delete is for a newly provisioned bkt vs an existing bkt
+	isNewBkt := r.internalClient.isNewBucketByOB(ob)
```
copejon: we're using `isNewBkt` and `isDynamicallyProvisioned` in naming to mean the same thing.  I lean toward `isDynamicallyProvisioned`

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -259,15 +266,18 @@ func (r *ObjectBucketClaimReconciler) handleDeleteClaim(key client.ObjectKey) er
...
 
-	class, err := storageClassForOB(ob, r.internalClient)
-	if err != nil || class == nil {
-		return fmt.Errorf("error getting storageclass from OB %q", ob.Name)
+	// Call the provisioner's `Revoke` method for old (brownfield) buckets regardless of reclaimPolicy.
+	// Also call `Revoke` for new buckets with a reclaimPolicy other than "Delete".
+	if ob.Spec.ReclaimPolicy == nil {
```
copejon: this is a pretty far out edge case.  worth checking?

jeffvance: remove check. Internal integrity err only and probably a panic is the right result of this ptr is nil.

PR #66, opened 2019-04-09
add linter to CI

copejon:


.travis.yml
```diff
@@ -11,15 +11,15 @@ go:
 before_script:
   - go version
   - export GOBIN="$GOPATH/bin"
-  - go get -u golang.org/x/lint/golint
   - go get golang.org/x/tools/cmd/goimports
+  - curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.16.0 && golangci-lint --version
```
jeffvance: Is there a potential issue with the version being hard-coded?

hack/go.sh
```diff
@@ -64,39 +65,76 @@ test(){
...
-    imports-check
-    vet
+    lint
     test
     build
 }
 
 help(){
-  msg=\
+  local msg=\
```
jeffvance: this is fine but you can also not set a var and simply:
`` 
`cat <<EOF
...
EOF
```

.travis.yml
copejon: hard coding the version is actually the recommended method (vs tracking head)

PR #67, opened 2019-04-09
Update OB and OBC phases

copejon:


pkg/provisioner/manager.go
```diff
@@ -16,6 +16,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
```
jeffvance: I may answer this later in the review, but is this an additional logger?

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -205,13 +210,23 @@ func (r *ObjectBucketClaimReconciler) handleProvisionClaim(key client.ObjectKey,
 	setObjectBucketName(ob, key)
 	ob.Spec.StorageClassName = obc.Spec.StorageClassName
 	ob.Spec.ClaimRef, err = claimRefForKey(key, r.internalClient)
-	ob.SetFinalizers([]string{finalizer})
 	ob.Spec.ReclaimPolicy = options.ReclaimPolicy
+	ob.SetFinalizers([]string{finalizer})
```
jeffvance: maybe in another pr the assignments to the ob, creation of the ob, and setting its status can become a helper func just to reduce the length of handleProvision.

pkg/provisioner/manager.go
copejon: this is the actual controller logger that we need to set if we want to get Reconile() errors reported

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -265,15 +280,24 @@ func (r *ObjectBucketClaimReconciler) handleDeleteClaim(key client.ObjectKey) er
 		return nil
 	}
 	// see if obc delete is for a newly provisioned bkt vs an existing bkt
-	isDynamicallyProvisioned := isNewBucketByOB(r.internalClient, ob)
 
 	// Call the provisioner's `Revoke` method for old (brownfield) buckets regardless of reclaimPolicy.
 	// Also call `Revoke` for new buckets with a reclaimPolicy other than "Delete".
 	if ob.Spec.ReclaimPolicy == nil {
```
jeffvance: de we need this check? I removed a similar check in handleProvision.

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -265,15 +280,24 @@ func (r *ObjectBucketClaimReconciler) handleDeleteClaim(key client.ObjectKey) er
...
 		log.Error(nil, "got null reclaimPolicy", "ob", ob.Name)
 		return nil
 	}
+
+	ob, err = updateObjectBucketPhase(r.internalClient, ob, v1alpha1.ObjectBucketClaimStatusPhaseReleased, r.retryInterval, r.retryTimeout)
```
jeffvance: no `err` check.

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -265,15 +280,24 @@ func (r *ObjectBucketClaimReconciler) handleDeleteClaim(key client.ObjectKey) er
...
+
+	class, err := storageClassForOB(ob, r.internalClient)
```
jeffvance: don't see `class` being used below.

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
```diff
@@ -205,13 +210,23 @@ func (r *ObjectBucketClaimReconciler) handleProvisionClaim(key client.ObjectKey,
...
 	ob.Spec.ClaimRef, err = claimRefForKey(key, r.internalClient)
-	ob.SetFinalizers([]string{finalizer})
 	ob.Spec.ReclaimPolicy = options.ReclaimPolicy
+	ob.SetFinalizers([]string{finalizer})
 
 	if ob, err = createObjectBucket(ob, r.internalClient, r.retryInterval, r.retryTimeout); err != nil {
 		return err
 	}
 
+	ob, err = updateObjectBucketPhase(r.internalClient, ob, v1alpha1.ObjectBucketStatusPhaseBound, r.retryInterval, r.retryTimeout)
```
jeffvance: why can't `CreateObjectBucket` above also set the status to be bound, vs. 2 API calls?

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
jeffvance: should there be a parallel _Released_ phase set for the OBC too, before its deleted?
_Released_ should be a very short-lived Phase. What is the benefit of setting this here and then (after `Delete` or `Revole`) deleting the ob?

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
copejon: probably a left over from a large rebase

pkg/provisioner/reconciler/claim-reconciler/resourcehandlers.go
```diff
@@ -238,3 +238,49 @@ func deleteObjectBucket(ob *v1alpha1.ObjectBucket, ic *internalClient) error {
 	}
 	return nil
 }
+
+func updateObjectBucketClaimPhase(ic *internalClient, obc *v1alpha1.ObjectBucketClaim, phase v1alpha1.ObjectBucketClaimStatusPhase, retryInterval, retryTimeout time.Duration) (*v1alpha1.ObjectBucketClaim, error) {
```
jeffvance: can all Phase updates for OBs and OBCs, be combined into a more generic func? Similar to `createUntilDefaultTimeout`?

pkg/provisioner/reconciler/claim-reconciler/resourcehandlers.go
```diff
@@ -238,3 +238,49 @@ func deleteObjectBucket(ob *v1alpha1.ObjectBucket, ic *internalClient) error {
...
+		}
+		return true, err
+	})
+	if err != nil {
+		return nil, fmt.Errorf("error updating phase: %v", err)
+	}
+	return ob, nil
+}
+
+func updateClaim(obc *v1alpha1.ObjectBucketClaim, c *internalClient) error {
```
jeffvance: should this be a `Poll` loop like above?

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
copejon: The OBC can't be updated because it's been deleted.  When we get the event, the API object is already gone.

Marking the OB as `released` gives us some indication an obc deletion occurred but has not completed.

pkg/provisioner/reconciler/claim-reconciler/reconiler.go
copejon: The `status` subresource can only be updated by the `client.UpdateStatus()` method.  Values in the OB.Status at `Create()` are ignored.

jeffvance: ok, I didn't know that...

jeffvance:
LGTM

PR #88, opened 2019-05-14
allow obc to have no bucket related fields for "brownfield"

jeffvance:
needed for brownfield bkts.


pkg/provisioner/controller.go
```diff
@@ -217,6 +217,10 @@ func (c *Controller) handleProvisionClaim(key string, obc *v1alpha1.ObjectBucket
 		err       error
 	)
 
+	if !shouldProvision(obc) {
```
jeffvance: @copejon I moved this line up. Pls confirm this is ok. My reason was to kick out sooner when we know we will...

copejon:
lgtm

PR #115, opened 2019-06-07
delete k8s resources after Delete/Revoke and set ownerReference for secret and CM

jeffvance:
Main purposes of this PR are:
1.  move k8s cleanup code to after the Delete() and Revoke() calls so that:
    -  certain provisioners (ceph) can access the generated secret in their Delete/Revoke methods.
    -  it's more obvious that the lib was unable to complete the Delete/Revoke.
1.  add ownerReferences to the generated CM and Secret and remove explicit delete calls from lib.

cc. @copejon


jeffvance:
Note: 1st commit implements moving cleanup code to after Delete/Revoke calls.
Later in this pr I'll add ownerReferences to the generated secret and CM

pkg/provisioner/controller.go
```diff
@@ -322,49 +322,35 @@ func (c *Controller) handleProvisionClaim(key string, obc *v1alpha1.ObjectBucket
...
 }
 
+// Delete or Revoke access to bucket defined by passed-in key.
 func (c *Controller) handleDeleteClaim(key string) error {
+	// Call `Delete` for new (greenfield) buckets with reclaimPolicy == "Delete".
+	// Call `Revoke` for new buckets with reclaimPolicy != "Delete".
+	// Call `Revoke` for existing (brownfield) buckets regardless of reclaimPolicy.
 
 	// TODO each delete should retry a few times to mitigate intermittent errors
+	// (Jon, is this TODO still needed??)
```
jeffvance: @copejon ?

jeffvance:
2nd commit adds `OwnerReferences` to the lib generated OB, Secret, and CM.

jeffvance:
TODO: check design doc to make sure OwnerReferences are shown and Delete/Revoke are describes correctl in terms of when k8s resources are deleted.  @jeffvance

pkg/provisioner/controller.go
copejon: yes, at the moment, the Delete() calls are a once off

pkg/provisioner/helpers.go
```diff
@@ -25,6 +25,19 @@ func makeObjectReference(claim *v1alpha1.ObjectBucketClaim) *corev1.ObjectRefere
         }
 }
 
+func makeOwnerReference(claim *v1alpha1.ObjectBucketClaim) metav1.OwnerReference {
+	blockOwnerDeletion := true
+	isController := true
+
+	return metav1.OwnerReference{
+		Kind: "ObjectBucketClaim",
```
copejon: - missing ApiVersion, Kind should not be hardcoded

getting it programatically looks like

```
   	groupVersion, kind := claim.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
        ...
            Kind: kind,
            ApiVersion: groupVersion,
        ...
```

pkg/provisioner/resourcehandlers.go
```diff
@@ -152,73 +158,57 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, c k
 	return configMap, err
 }
 
+// Only the finalizer needs to be removed. The CM will be deleted since it's
+// ownerReference refers to the parent OBC.
 func deleteConfigMap(cm *corev1.ConfigMap, c kubernetes.Interface) error {
```
copejon: should probably be renamed since it no longer deletes the CM.  Maybe `releaseConfigMap` or something

pkg/provisioner/resourcehandlers.go
```diff
@@ -152,73 +158,57 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, c k
...
 	if cm == nil {
 		log.Info("got nil configMap pointer, skipping delete")
 		return nil
 	}
+
+	logD.Info("ConfigMap is automatically deleted after its finalizer is removed", "name", cm.Namespace+"/"+cm.Name)
```
copejon: more precise to say `allowing configMap to be garbage collected`.

pkg/provisioner/resourcehandlers.go
```diff
@@ -152,73 +158,57 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, c k
...
-			return nil
-		}
-		return fmt.Errorf("error deleting configMap %s/%s: %v", cm.Namespace, cm.Name, err)
-	}
 	return nil
 }
 
+// Only the finalizer needs to be removed. The Secret will be deleted since it's
+// ownerReference refers to the parent OBC.
 func deleteSecret(sec *corev1.Secret, c kubernetes.Interface) error {
```
copejon: ditto rename

pkg/provisioner/resourcehandlers.go
```diff
@@ -152,73 +158,57 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, c k
...
 	if sec == nil {
 		log.Info("got nil secret, skipping")
 		return nil
 	}
 
+	logD.Info("secret is automatically deleted after its finalizer is removed", "name", sec.Namespace+"/"+sec.Name)
```
copejon: ditto garbage collected comment

pkg/provisioner/resourcehandlers.go
```diff
@@ -152,73 +158,57 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, c k
...
-	logD.Info("deleting ObjectBucket", "name", ob.Name)
+
+	logD.Info("OB is automatically deleted after its finalizer is removed", "name", ob.Namespace+"/"+ob.Name)
 	removeFinalizer(ob)
 	ob, err := c.ObjectbucketV1alpha1().ObjectBuckets().Update(ob)
 	if err != nil {
 		return err
 	}
 
-	err = c.ObjectbucketV1alpha1().ObjectBuckets().Delete(ob.Name, &metav1.DeleteOptions{})
```
copejon: The OB is not garbage collected since they cannot have an owner reference to a namespace-scoped object.

> Note: Cross-namespace owner references is disallowed by design. This means: 1) Namespace-scoped dependents can only specify owners in the same namespace, and owners that are cluster-scoped. **2) Cluster-scoped dependents can only specify cluster-scoped owners, but not namespace-scoped owners.**

[reference](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents)

This code shouldn't be removed as the OB must be manually deleted.

pkg/provisioner/resourcehandlers.go
```diff
@@ -152,73 +158,57 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, c k
...
+// Only the finalizer needs to be removed. The OB will be deleted since it's ownerReference
+// refers to the parent OBC.
 func deleteObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface) error {
 	if ob == nil {
 		log.Error(fmt.Errorf("got nil objectBucket, skipping"), "")
 		return nil
 	}
-	logD.Info("deleting ObjectBucket", "name", ob.Name)
+
+	logD.Info("OB is automatically deleted after its finalizer is removed", "name", ob.Namespace+"/"+ob.Name)
```
copejon: should revert, per below

pkg/provisioner/resourcehandlers.go
```diff
@@ -152,73 +158,57 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, c k
...
-		if errors.IsNotFound(err) {
-			log.Error(err, "secret vanished before we could delete it, skipping")
-			return nil
-		}
-		return fmt.Errorf("error deleting Secret %s/%s: %v", sec.Namespace, sec.Name, err)
-	}
 	return nil
 }
 
+// Only the finalizer needs to be removed. The OB will be deleted since it's ownerReference
```
copejon: revert, per below

pkg/provisioner/controller.go
```diff
@@ -389,7 +376,9 @@ func (c *Controller) handleDeleteClaim(key string) error {
 			return fmt.Errorf("error deleting objectBucket %v", ob.Name)
 		}
 	}
-	return nil
+
+	// delete all generated kubernetes resources
+	return c.deleteResources(ob, cm, secret)
```
copejon: the delete on the OB is redundant with line 372.  The reason the 2nd delete call doesn't fail is that we ignore NotFound errors.  The result is an inaccurate log statement

jeffvance: Good point! I'll delete L372-378 and let `deleteResources` do all the deletes.

pkg/provisioner/resourcehandlers.go
```diff
@@ -66,8 +71,11 @@ func newBucketConfigMap(ep *v1alpha1.Endpoint, obc *v1alpha1.ObjectBucketClaim)
 	}, nil
 }
 
-// NewCredentialsSecret returns a secret with data appropriate to the supported authenticaion method. Even if the values
-// for the Authentication keys are empty, we generate the secret.
+// NewCredentialsSecret returns a secret with data appropriate to the supported authenticaion
```
copejon: nit - func name should match case of function.  this will mess up GoDocs

PR #122, opened 2019-06-27
call UpdateStatus to update phase

jeffvance:

### DIFF
--- a/pkg/provisioner/resourcehandlers.go
+++ b/pkg/provisioner/resourcehandlers.go
@@ -102,23 +102,20 @@ func newCredentialsSecret(obc *v1alpha1.ObjectBucketClaim, auth *v1alpha1.Authen
 
 // createObjectBucket creates an OB based on the passed-in ob spec.
 // Note: a finalizer has been added to reduce chances of the ob being accidentally deleted.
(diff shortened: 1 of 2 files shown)


pkg/provisioner/resourcehandlers.go
```diff
@@ -226,40 +226,36 @@ func deleteObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface) error
 	return nil
 }
 
-func updateClaim(c versioned.Interface, obc *v1alpha1.ObjectBucketClaim, retryInterval, retryTimeout time.Duration) (*v1alpha1.ObjectBucketClaim, error) {
-	err := wait.PollImmediate(retryInterval, retryTimeout, func() (done bool, err error) {
+func updateClaim(c versioned.Interface, obc *v1alpha1.ObjectBucketClaim, retryInterval, retryTimeout time.Duration) (result *v1alpha1.ObjectBucketClaim, err error) {
```
copejon: named return `result` is unused, so a nil obc is always returned.

jeffvance: good catch. A c&p error on my part. The code still works due to pointers being used but it is a bug.

PR #124, opened 2019-07-01
Finalizer Removal Bug Fix & Logging clean up

copejon:


pkg/provisioner/resourcehandlers.go
```diff
@@ -235,31 +246,38 @@ func updateClaim(c versioned.Interface, obc *v1alpha1.ObjectBucketClaim, retryIn
...
 	}
 	return obc, nil
 }
 
 func updateObjectBucketClaimPhase(c versioned.Interface, obc *v1alpha1.ObjectBucketClaim, phase v1alpha1.ObjectBucketClaimStatusPhase, retryInterval, retryTimeout time.Duration) (*v1alpha1.ObjectBucketClaim, error) {
 	obc.Status.Phase = phase
-	obc, err := updateClaim(c, obc, retryInterval, retryTimeout)
+	err := wait.PollImmediate(retryInterval, retryTimeout, func() (done bool, err error) {
+		obc, err = c.ObjectbucketV1alpha1().ObjectBucketClaims(obc.Namespace).UpdateStatus(obc)
+		if err != nil {
```
jeffvance: nit and I can simplify in pr 122:  
```
   obc, err = c.ObjectbucketV1alpha1().ObjectBucketClaims(obc.Namespace).UpdateStatus(obc)
   return (err == nil), err
```

jeffvance:
lgtm
There's some overlap with #122 where `UpdateStatus` is now called and with some err msg refactoring.

pkg/provisioner/helpers.go
```diff
@@ -151,10 +147,9 @@ func composeBucketName(obc *v1alpha1.ObjectBucketClaim) (string, error) {
 	}
 	bucketName := obc.Spec.BucketName
 	if bucketName == "" {
-		logD.Info("bucket name is empty, generating...")
 		bucketName = generateBucketName(obc.Spec.GenerateBucketName)
-		logD.Info("generated bucket name", "name", bucketName)
 	}
+	logD.Info("generated bucket name", "name", bucketName)
```
jeffvance: The bucket name is not always generated. I think this change should be reverted...

pkg/provisioner/helpers.go
```diff
@@ -209,8 +203,8 @@ func removeFinalizer(obj metav1.Object) {
 	finalizers := obj.GetFinalizers()
 	for i, f := range finalizers {
 		if f == finalizer {
-			logD.Info("removing finalizer", "object", obj.GetName(), "finalizer", finalizer)
 			obj.SetFinalizers(append(finalizers[:i], finalizers[i+1:]...))
+			logD.Info("finalizer removed")
```
jeffvance: can we get the k8s object name in this msg so we see which finalizer was removed. Or does that matter?
Can `SetFinalizer` ever return an error cuz if so then I'd prefer to *only* log an error here...

pkg/provisioner/manager.go
```diff
@@ -79,7 +79,7 @@ func NewProvisioner(
 // Run starts the claim and bucket controllers.
 func (p *Provisioner) Run(stopCh <-chan struct{}) (err error) {
 	defer klog.Flush()
-	log.Info("starting provisioner Controller")
+	log.Info("starting provisioner", "name", p.Name)
```
jeffvance: +1

pkg/provisioner/resourcehandlers.go
```diff
@@ -123,7 +119,7 @@ func createSecret(obc *v1alpha1.ObjectBucketClaim, auth *v1alpha1.Authentication
 	if err != nil {
 		return nil, err
 	}
-
+	logD.Info("creating Secret", "name", secret.Namespace+"/"+secret.Name)
```
jeffvance: super nit: since this pr makes logging msgs more consistent (esp in terms of upper vs. lowercase) should Secret be capitalized here?

pkg/provisioner/resourcehandlers.go
```diff
@@ -146,6 +142,7 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, c k
 		return nil, err
 	}
 
+	logD.Info("creating ConfigMap", "name", configMap.Namespace+"/"+configMap.Name)
```
jeffvance: ditto

pkg/provisioner/resourcehandlers.go
```diff
@@ -164,14 +161,18 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, c k
...
 // Only the finalizer needs to be removed. The CM will be garbage collected since its
 // ownerReference refers to the parent OBC.
-func releaseConfigMap(cm *corev1.ConfigMap, c kubernetes.Interface) error {
+func releaseConfigMap(cm *corev1.ConfigMap, c kubernetes.Interface) (err error) {
 	if cm == nil {
 		return nil
 	}
 
-	logD.Info("ConfigMap is garbage collected after its finalizer is removed", "name", cm.Namespace+"/"+cm.Name)
+	cm, err = c.CoreV1().ConfigMaps(cm.Namespace).Get(cm.Name, metav1.GetOptions{})
```
jeffvance: Is the reason for this because when the finalizer is removed our copy of the CM is stale and thus the `Update` below will fail?

jeffvance: I thought this would be needed so that the Update below is using a fresh copy of the object? Why aren't we re-Getting the CM after the finalizer is removed?

pkg/provisioner/resourcehandlers.go
```diff
@@ -164,14 +161,18 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, c k
...
+	if err != nil {
+		return err
+	}
+	logD.Info("removing ConfigMaps's finalizer", "name", cm.Namespace+"/"+cm.Name)
 	removeFinalizer(cm)
```
jeffvance: Possibly `removeFinalizer` should return the object?

pkg/provisioner/resourcehandlers.go
```diff
@@ -181,15 +182,18 @@ func releaseConfigMap(cm *corev1.ConfigMap, c kubernetes.Interface) error {
...
 // ownerReference refers to the parent OBC.
-func releaseSecret(sec *corev1.Secret, c kubernetes.Interface) error {
+func releaseSecret(sec *corev1.Secret, c kubernetes.Interface) (err error) {
 	if sec == nil {
 		log.Info("got nil secret, skipping")
 		return nil
 	}
-
-	logD.Info("secret is garbage collected after its finalizer is removed", "name", sec.Namespace+"/"+sec.Name)
+	sec, err = c.CoreV1().Secrets(sec.Namespace).Get(sec.Name, metav1.GetOptions{})
```
jeffvance: same comment as for CM above...

pkg/provisioner/resourcehandlers.go
```diff
@@ -199,27 +203,30 @@ func releaseSecret(sec *corev1.Secret, c kubernetes.Interface) error {
...
+// https://github.com/kubernetes/kubernetes/issues/50037
 func deleteObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface) error {
 	if ob == nil {
 		return nil
 	}
 
-	logD.Info("deleting OB after its finalizer is removed", "name", ob.Name)
+	logD.Info("removing ObjectBucket's finalizer", "name", ob.Name)
 	removeFinalizer(ob)
 	ob, err := c.ObjectbucketV1alpha1().ObjectBuckets().Update(ob)
```
jeffvance: why isn't the OB stale here too and thus the Update fails?

pkg/provisioner/resourcehandlers.go
copejon: the object is passed by ref, so no need

pkg/provisioner/resourcehandlers.go
copejon: the object is passed by ref, so no need

jeffvance:
lgtm

PR #136, opened 2019-08-23
Bump kubernetes version to 14.1

rohantmp:
Rook is being bumped to 1.14.1, but it can't update until this is bumped to 1.14.1 as it imports this.

I had previously said:
> Can we please merge this into a 1.14.1 branch so that merging this doesn't break rook  and then when rook updates it can point to the 1.14.1 branch.
> 
> That is the only safe update strategy I can think of...
> 
> @travisn @leseb @copejon 

This turned out to nonsense and handled by Gopkg.lock

Signed-off-by: Rohan CJ <rohantmp@gmail.com>

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -10,16 +10,16 @@ required = [
 
 [[override]]
   name = "k8s.io/code-generator"
-  version = "kubernetes-1.14.0"
+  version = "kubernetes-1.14.1"
 
 [[constraint]]
   name = "k8s.io/api"
-  version = "kubernetes-1.14.0"
+  version = "kubernetes-1.14.1"
 
 [[constraint]]
   name = "k8s.io/apimachinery"
-  version = "kubernetes-1.14.0"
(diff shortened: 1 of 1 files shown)


rohantmp:
Please merge this

guymguym:
@rohantmp We want to support OCP 3.11. Does this prevent it?

guymguym:
Oh, just noticed that this was already 1.14.0 before your PR.
But will that be a problem on OCP 3 (kube 1.13)? 
I don't know if this library uses any API's that are specific to 1.14 and not 1.13...

rohantmp:
I think that it will be fine based on: https://kubernetes.io/docs/setup/release/version-skew-policy/

Maybe one thing to keep an eye out for would be deprecated api versions (Eg: alpha of some resource no longer available), but that won't be a problem as far as updating the libraries goes.

rohantmp:
@guymguym can we merge this?

guymguym:
Yes of course!

Issue #57, opened 2019-04-04
consider adding labels to OB, Secret, ConfigMap based on provisioner name


jeffvance:
Per original design doc:
```
kind: Secret
metadata:
  labels:
    objectbucket.io/PROVISIONER-NAME: [3]
...
1. label may be used to associate all artifacts under a particular provisioner.
```

```
kind: ConfigMap
metadata:
  labels:
    objectbucket.io/PROVISIONER-NAME: [3]
...
1. label here associates all artifacts under a spoecific provisioner.
```

```
kind: ObjectBucket
metadata:
  labels:
    objectbucket.io/PROVISIONER-NAME: [3]
...
1. label here associates all artifacts under the particular provisioner.
```

PR #140, opened 2019-08-28
Add resource labels

### LINKED ISSUE #133: Add provisioner label to OB,OBC,Secret,ConfigMap for automated cleanup
The use case is when uninstalling a provisioner that is not running, we need a way to clean all of its resources.
Since all these resources OB,OBC,Secret,ConfigMap are not confined to any namespace, it would be best to add a label `provisioner=<provisioner-name>` to all of them.

copejon:
Fixes #133 #57 

Adds a label (`<provisioner_name>: ""`) to the OBC, OB, CM, and Secret.  By k8s convention, provisioner names are usually in the form of `qualified.name/foo`.   Labels _values_ may only have `.`, `_`, or `-` special characters whereas label _keys_ are parsed in the same manner as provisioner names.  Thus, assigning the name to the label key does not create an intuitive experience when getting object by label.

The method for getting by label is the same: `kubectl get <obj> -l provisionerName`

A little refactoring was done on CreateSecret and CreateConfigMap to separate process concerns (defining the object vs creating the object) and to clean up function signatures.

**TODO:** adjust unit tests


jeffvance:
LGTM

pkg/provisioner/controller.go
```diff
@@ -220,12 +221,18 @@ func (c *Controller) syncHandler(key string) error {
 func (c *Controller) handleProvisionClaim(key string, obc *v1alpha1.ObjectBucketClaim, class *storagev1.StorageClass) (err error) {
 
 	var (
-		ob        *v1alpha1.ObjectBucket
-		secret    *corev1.Secret
-		configMap *corev1.ConfigMap
+		ob               *v1alpha1.ObjectBucket
+		secret           *corev1.Secret
+		configMap        *corev1.ConfigMap
+		provisionerLabel = map[string]string{c.provisionerName: ""}
```
guymguym: I wouldn't use empty string value. 
That's too confusing for no good reason.
How about the value will be provided by the operator that initializes the library?
I think it makes sense that each operator will use its version as the label value so that we can identify which version of the operator created those?

pkg/provisioner/controller.go
```diff
@@ -220,12 +221,18 @@ func (c *Controller) syncHandler(key string) error {
...
 	)
 	obcNsName := obc.Namespace + "/" + obc.Name
 
+	obc.ObjectMeta.SetLabels(provisionerLabel)
```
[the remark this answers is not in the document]
guymguym: @jeffvance How much time can an OBC be in memory before this?
Is that OBC coming from the watch event cache?
I didn't read the entire code path but maybe you remember already and can tell...

pkg/provisioner/controller.go
```diff
@@ -293,24 +300,34 @@ func (c *Controller) handleProvisionClaim(key string, obc *v1alpha1.ObjectBucket
 		return fmt.Errorf("error %s bucket: %v", verb, err)
 	} else if ob == (&v1alpha1.ObjectBucket{}) {
 		return fmt.Errorf("provisioner returned nil/empty object bucket")
+	} else if err := validateObjectBucket(ob); err != nil {
```
guymguym: should be`=` instead of `:=` if we want the deferred delete-on-error to run in this case or else the outer err will stay nil

pkg/provisioner/controller.go
```diff
@@ -220,12 +221,18 @@ func (c *Controller) syncHandler(key string) error {
...
-		configMap *corev1.ConfigMap
+		ob               *v1alpha1.ObjectBucket
+		secret           *corev1.Secret
+		configMap        *corev1.ConfigMap
+		provisionerLabel = map[string]string{c.provisionerName: ""}
 	)
 	obcNsName := obc.Namespace + "/" + obc.Name
 
+	obc.ObjectMeta.SetLabels(provisionerLabel)
+	if obc, err = c.libClientset.ObjectbucketV1alpha1().ObjectBucketClaims(obc.Namespace).Update(obc); err != nil {
```
guymguym: I wish this update can be called once for the combined update of the labels, the new finalizer ([protectOBC](https://github.com/kube-object-storage/lib-bucket-provisioner/pull/139/files#diff-5d07bb1605a6a9c2d1308ba7348be5f3R253) from #139 and the phase update to "pending".

pkg/provisioner/resourcehandlers.go
```diff
@@ -279,4 +267,15 @@ func updateObjectBucketPhase(c versioned.Interface, ob *v1alpha1.ObjectBucket, p
...
+
+func validateObjectBucket(ob *v1alpha1.ObjectBucket) error {
+	var errs []string
+	if ob.Spec.Endpoint == nil {
+		errs = append(errs, "Spec.Endpoint cannot be nil")
+	}
+	if ob.Spec.Authentication == nil {
+		errs = append(errs, "Spec.Authentication cannot be nil")
+	}
+	return fmt.Errorf("%s", strings.Join(errs, "; "))
```
guymguym: Seems like this function returns a non nil error even if errs array is empty...
Did you mean this:
```suggestion
	if len(errs) > 0 {
		return fmt.Errorf("%s", strings.Join(errs, "; "))
	}
	return nil
```

pkg/provisioner/controller.go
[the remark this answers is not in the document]
copejon: The OBC is gotten for the first time on line 177 via API call.  I'll see if i can expose a race condition with some manually looping tests.

pkg/provisioner/controller.go
copejon: Unfortunately, we can only get it down to a min of 2 calls.  UpdateStatus is exclusive to object status updates whereas Update only applies changes to Spec and Metadata.   We can condense the finalizer and label updates once this PR or #139 is merged, then make the changes in the other.

pkg/provisioner/resourcehandlers.go
copejon: Right ... wonder why this hasn't triggered in testing.

pkg/provisioner/controller.go
jeffvance: The provisioner already supplies their name in the `NewProvisione` lib call so perhaps this can be used as the value here?

guymguym: the name is useless - the version will help troubleshooting much more.

jeffvance: Are these labels for trouble shooting or making cleanup simpler?  Version is probably useful but is not currently passed to the library.

guymguym: Both. So lets add another ctor param for label value.

copejon: @jeffvance 
> The provisioner already supplies their name in the NewProvisioner lib call so perhaps this can be used as the value here?

see PR description of why we have to use it as the key and not the value

@guymguym 
> the name is useless - the version will help troubleshooting much more.

I don't really agree - obcs in particular don't identify their provisioner.  Labeling our resources allows for getting them by their provisioner, which I definitely useful.  e.g.
`kubectl get -l <provisionerName> obc,secret,cm --all-namespaces` or
`kubectl get ob -l <provisionerName>`

This works even if the value is empty.

> That's too confusing for no good reason.

I'm not following this, why would it be confusing?

guymguym: An empty string value is confusing. Certainly in go.

The name is only useless as a label value, since the key is already the provisioner name...

I thought what is the most helpful value, and the operator version is the best I came up with. Do you see a better option?

pkg/provisioner/controller.go
```diff
@@ -53,24 +52,26 @@ type Controller struct {
...
-	provisionerName string
+	provisioner        api.Provisioner
+	provisionerName    string
+	provisionerVersion string
 }
 
 var _ controller = &Controller{}
 
-func NewController(provisionerName string, provisioner api.Provisioner, clientset kubernetes.Interface, crdClientSet versioned.Interface, obcInformer informers.ObjectBucketClaimInformer, obInformer informers.ObjectBucketInformer) *Controller {
+func NewController(provisionerName string, operatorVersion string, provisioner api.Provisioner, clientset kubernetes.Interface, crdClientSet versioned.Interface, obcInformer informers.ObjectBucketClaimInformer, obInformer informers.ObjectBucketInformer) *Controller {
```
jeffvance: This is a breaking api change. Is there a way to make version optional? Otherwise don't we need to bump the lib version?

pkg/provisioner/resourcehandlers.go
```diff
@@ -292,13 +280,33 @@ func updateObjectBucketClaimPhase(c versioned.Interface, obc *v1alpha1.ObjectBuc
...
+	if ob.Spec.Authentication == nil {
+		errs = append(errs, "Spec.Authentication cannot be nil")
+	}
+	if len(errs) > 0 {
+		return fmt.Errorf("%s", strings.Join(errs, "; "))
+	}
+	return nil
+}
+
+func setClaimOwnership(obc *v1alpha1.ObjectBucketClaim, labels map[string]string, finalizers []string) *v1alpha1.ObjectBucketClaim {
```
jeffvance: maybe func name should be more generic? The name implies only ownership is set but labels are too.

pkg/provisioner/controller.go
```diff
@@ -238,20 +239,20 @@ func (c *Controller) syncHandler(key string) error {
...
 	if obc, err = updateObjectBucketClaimPhase(c.libClientset, obc, v1alpha1.ObjectBucketClaimStatusPhasePending, defaultRetryBaseInterval, defaultRetryTimeout); err != nil {
-		return fmt.Errorf("error updating OBC %q's status to %q: %v", obcNsName, v1alpha1.ObjectBucketClaimStatusPhasePending, err)
+		return fmt.Errorf("error updating OBC %q's status to %q: %v", key, v1alpha1.ObjectBucketClaimStatusPhasePending, err)
 	}
 
-	// set finalizer in OBC so that resources can be cleaned up when the obc is deleted
-	if err = c.protectOBC(obc); err != nil {
-		return err
+	setClaimOwnership(obc, provisionerLabel, []string{finalizer})
+	if obc, err = updateClaim(c.libClientset, obc, defaultRetryBaseInterval, defaultRetryTimeout); err != nil {
```
jeffvance: the error is updating the obc, not necessarily just labels...

pkg/provisioner/controller.go
```diff
@@ -238,20 +239,20 @@ func (c *Controller) syncHandler(key string) error {
...
+		return fmt.Errorf("error setting provisioner label")
```
jeffvance: the error is updating the obc, not necessarily just labels...

pkg/provisioner/resourcehandlers.go
```diff
@@ -274,14 +262,14 @@ func updateClaim(c versioned.Interface, obc *v1alpha1.ObjectBucketClaim, retryIn
...
-		return (err == nil), err
+		return err == nil, err
 	})
 	return
 }
 
 func updateObjectBucketClaimPhase(c versioned.Interface, obc *v1alpha1.ObjectBucketClaim, phase v1alpha1.ObjectBucketClaimStatusPhase, retryInterval, retryTimeout time.Duration) (result *v1alpha1.ObjectBucketClaim, err error) {
-	logD.Info("updating status:", "obc", obc.Namespace+"/"+obc.Name, "old status",
-		obc.Status.Phase, "new status", phase)
+	logD.Info("updating objectBucketClaim status:", "current phase",
```
jeffvance: the obc name is not needed in the log?

pkg/provisioner/controller.go
guymguym: Where is the lib version?

copejon:
@guymguym So in adding a setable version value, we'll be adding a new param to NewProvisioner().  What do you think about making that a separate PR, given it's an API breaking change?  This PR would still add basic labeling with the provisioner name as the key.

jeffvance:
Summary of discussion with @guymguym:
The lib's `Provisioner` struct can have a label-value string field added. This becomes the value of `<provisioner-name>=<label-value>` label with the default value being "".

copejon:
closing in favor of #159

PR #141, opened 2019-08-29
CRD's additionalPrinterColumns for kubectl get

guymguym:
Defined the additionalPrinterColumns using kubebuilder marker comments, see:
https://book.kubebuilder.io/reference/generating-crd.html

However, since the CRD's of the lib are not generated by kubebuilder, I manually applied this change to the crds file.

### DIFF
--- a/deploy/customResourceDefinitions.yaml
+++ b/deploy/customResourceDefinitions.yaml
@@ -19,6 +19,18 @@ spec:
   scope: Cluster
   subresources:
     status: {}
+  additionalPrinterColumns:
+  - JSONPath: .spec.storageClassName
+    description: StorageClass
+    name: StorageClass
+    type: string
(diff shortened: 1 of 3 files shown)


deploy/customResourceDefinitions.yaml
```diff
@@ -41,3 +53,27 @@ spec:
   scope: Namespaced
   subresources:
     status: {}
+  additionalPrinterColumns:
```
jeffvance: glad to see you're not bothering with the unused obc fields, Versioned, SSL, etc.

jeffvance:
LGTM

deploy/customResourceDefinitions.yaml
guymguym: I am bothered that it is not generated the go structures.

PR #144, opened 2019-08-31
TravisCI

guymguym:


guymguym:
@copejon @jeffvance  `go test ./pkg/...` is failing on CI. Did it work for you?

guymguym:
Another question - I saw you listed Gopkg.lock in gitignore.
I don't think this is the recommended way. Was there a problem with it?

copejon:
> Another question - I saw you listed Gopkg.lock in gitignore.

I think this is an artifact of dependency resolution from a while back.  Re-reading dep docs, it still says that Gopkg.lock is ignored in dependencies of projects (i.e. the lib), so un-ignoring it should be safe.

> go test ./pkg/... is failing on CI. Did it work for you?

The tests are stale.  The ownerRef is causing the failures here - the `want` object leaves it nil but the returned OBC contains one.   I already addressed this  #140 , will extract those changes and open a PR to this branch so it'll go in with the CI fixes.

jeffvance:
So should this pr also not hide _Gopkgg.lock?

.travis.yml
```diff
@@ -1,31 +1,10 @@
...
-  - curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
-
 install:
-  - dep ensure -v
-
-before_script:
-  - go version
-  - export GOBIN="$GOPATH/bin"
-  - go get golang.org/x/tools/cmd/goimports
-  - curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.16.0 && golangci-lint --version
```
copejon: golangci-lint is an industry standard go linter aggregator and offers far more configurability and reliability than golint.  In addition to basic linting, it's configured to perform a number of other useful checks.  The full list of enabled and disabled linters can be seen by `./hack/go.sh linters`.  Executing the linters can be done with `./hack/go.sh lint`.

Makefile
```diff
@@ -0,0 +1,46 @@
...
+.PHONY: vet
+
+lint: vendor
+	go get -u golang.org/x/lint/golint
+	golint -set_exit_status=1 ./pkg/...
+	@echo "lint - done."
+.PHONY: lint
+
+test: vendor
+	go test ./pkg/...
```
copejon: Testing ./pkg/... is problematic because of some issues with generated k8s code.  As a result, fine tuning is required to selective test only the human written project files.  use `./hack/go.sh test` instead

Makefile
```diff
@@ -0,0 +1,46 @@
+all: build fmt imports vet lint test
```
copejon: I'm realizing that we never documented our build script in the ./hack/README.md.  `./hack/go.sh` handles most of these operations, excluding 'vendor'.  

The script handles some known issues with testing (see below), runs a preconfigured `go imports` so we always structure imports uniformly between devs, executes robust linting, and wraps up a combination of these for CI to run.  This also means devs can run CI workflow locally to get a quick glance on whether it will pass.  Obviously this last part isn't a perfect recreation of CI as env's vary, but it's is still useful.

.travis.yml
```diff
@@ -1,31 +1,10 @@
----
-sudo: false
-
 os: linux
-
+dist: xenial
+sudo: required
```
copejon: Why is sudo required?

.travis.yml
copejon: The config file can be found at [.golangci.yaml](https://github.com/kube-object-storage/lib-bucket-provisioner/blob/master/.golangci.yml)

.travis.yml
```diff
@@ -1,31 +1,10 @@
...
-  - go get golang.org/x/tools/cmd/goimports
-  - curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.16.0 && golangci-lint --version
-
+  - curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
 script:
-  - ./hack/go.sh ci-checks
-
-deploy:
-  - provider: script
-    script: ./hack/go.sh ci-checks
```
copejon: As stated elsewhere - `./hack/go.sh ci-checks` runs a bespoke set of build/vet/lint/test commands for the project.  Unless there's good reason not to use it, I would prefer to stick with this script.

.travis.yml
```diff
@@ -1,31 +1,10 @@
...
-    on:
-      branch: master
```
copejon: Is there a reason to build on all commits for working branches in the repo?  If later we add release-branches/versioning, then I think we would update this list.  Otherwise, we can cut down job spam in Travis by limiting to master.  PRs always run in CI.

PR #150, opened 2019-09-06
remove string concat

copejon:
Removes reference to unchecked ptr.  Concatenation is unnecessary as the key (obcNamespace/obcName) is embedded in the logger.


jeffvance:
lgtm

PR #157, opened 2019-09-09
disable golangci-ling go mod support

copejon:
Golangci-lint will mistakenly build the project using local mods of dependencies if present.  Disable this behavior to force it to build from vendor.


jeffvance:
lgtm. Merge when CI done.

PR #159, opened 2019-09-10
Custom labels

### LINKED ISSUE #133: Add provisioner label to OB,OBC,Secret,ConfigMap for automated cleanup

copejon:
This PR introduces a backwards compatible change to enable the setting of labels on resources managed by the provisioner.  This is accomplished by adding a`SetLabels(map[string]string)` method to the `Provisioner` interface.  The method statically assigns the map to the controller for access during sync operations.

The purpose of this change is to ease api resource management and provisioner development. 

Also include some minor refactoring for readability.

fixes #133, #57 and replaces pr #140 

### DIFF
--- a/pkg/provisioner/controller.go
+++ b/pkg/provisioner/controller.go
@@ -249,7 +249,6 @@ func (c *Controller) handleProvisionClaim(key string, obc *v1alpha1.ObjectBucket
 		secret    *corev1.Secret
 		configMap *corev1.ConfigMap
 	)
-	obcNsName := obc.Namespace + "/" + obc.Name
 
 	// first step is to update the OBC's status to pending
 	obc, err = updateObjectBucketClaimPhase(
@@ -259,7 +258,7 @@ func (c *Controller) handleProvisionClaim(key string, obc *v1alpha1.ObjectBucket
 		defaultRetryBaseInterval,
 		defaultRetryTimeout)
 	if err != nil {
-		return fmt.Errorf("error updating OBC %q's status to %q: %v", obcNsName, v1alpha1.ObjectBucketClaimStatusPhasePending, err)
+		return fmt.Errorf("error updating OBC status to %q: %v", v1alpha1.ObjectBucketClaimStatusPhasePending, err)
 	}
 
 	// set finalizer in OBC so that resources can be cleaned up when the obc is deleted
@@ -340,7 +339,7 @@ func (c *Controller) handleProvisionClaim(key string, obc *v1alpha1.ObjectBucket
 		defaultRetryBaseInterval,
 		defaultRetryTimeout)
 	if err != nil {
-		return fmt.Errorf("error creating secret for OBC %q: %v", obcNsName, err)
+		return fmt.Errorf("error creating secret for OBC: %v", err)
 	}
 	configMap, err = createConfigMap(
 		obc,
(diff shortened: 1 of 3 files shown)


pkg/provisioner/controller.go
```diff
@@ -40,6 +40,7 @@ import (
 
 type controller interface {
 	Start(<-chan struct{}) error
+	SetLabels(map[string]string)
 }
 
 // Provisioner is a CRD Controller responsible for executing the Reconcile() function in response to OB and OBC events.
```
jeffvance: Can you change this comment to not mention  OB events?

pkg/provisioner/controller.go
```diff
@@ -53,6 +54,8 @@ type Controller struct {
 	obHasSynced  cache.InformerSynced
 	queue        workqueue.RateLimitingInterface
 
+	provisionerLabels map[string]string
```
jeffvance: can you add a comment saying these labels are optional and defined completely by each provisioner?

jeffvance: nit: not sure if there's consistency with the use of blank lines in this struct. I'd prefer a comment before each field and no blank lines.

pkg/provisioner/controller.go
```diff
@@ -124,6 +117,20 @@ func (c *Controller) Start(stopCh <-chan struct{}) error {
 	return nil
 }
 
+func (c *Controller) SetLabels(labels map[string]string) {
```
jeffvance: It's a little inconsistent that we have a setter for labels but not for any of the other fields in the Controller struct. I know the reason is to preserve backwards compatibility, but a year from now this may seem odd. I think the only other setter we might want is provisioner name but that's also pretty obvious to pass to `NewProvisioner`... so this is just a rambling comment....

pkg/provisioner/controller.go
```diff
@@ -464,21 +513,24 @@ func (c *Controller) deleteResources(ob *v1alpha1.ObjectBucket, cm *corev1.Confi
...
+func (c *Controller) setOBCMetaFields(obc *v1alpha1.ObjectBucketClaim) (err error) {
 
 	clib := c.libClientset
-	obcNsName := obc.Namespace + "/" + obc.Name
 
+	logD.Info("getting OBC to set metadata fields")
 	obc, err = clib.ObjectbucketV1alpha1().ObjectBucketClaims(obc.Namespace).Get(obc.Name, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("unable to Get obc %q for protection: %v", obcNsName, err)
+		return fmt.Errorf("error getting unconfigured obc: %v", err)
```
jeffvance: what does "unconfigured" mean? May be better to omit?

pkg/provisioner/controller.go
copejon: I agree but I think it's out of the scope of this PR.  Labels have had significant discussion - I don't want to bog this code down with conversation around custom annotations, finalizers, etc.

pkg/provisioner/controller.go
copejon: fair point - I was struggling to think of how to describe these yet-unclaimed OBCs.

jeffvance:
Need to update s3 and ceph provisioner docs to include the `SetLabels` call. 

**Note**: the 2 sample provisioner code links [here](https://github.com/kube-object-storage/lib-bucket-provisioner/blob/master/doc/examples/sample-how-to-write-provisioner.md#sample-code) are dead.

jeffvance:
Quick testing shows this works as described. I changed the s3 provisioner to call `SetLabels` and I see the label on the OB, OBC, Secret and CM.

LGTM

PR #160, opened 2019-09-10
Allow for namespace scoped provisioner

copejon:
Honor the `NewProvisioner()` parameter list by making use of the `namespace` parameter if provided.

If no namespace is given, the informers are generated as cluster scoped, as they have been.  If a namespace is provided, the informers are scoped to that namespace.  The result of course is that the controller will only monitor OBCs in the given namespace.

This change is backward compatible.


jeffvance:
lgtm

PR #166, opened 2019-09-12
set obc status earlier

### LINKED ISSUE #158: obc status is not set when there is an early failure
the OBC status is not updated if the provisioning flow fails early.

e.g. try to use a nonexisting storage class. the provisioner will fail, but status is not set on the obc

jeffvance:
fixes #158 
Error path tested.

@copejon @dannyzaken


copejon:
If an error then occurs, e.g. the SC doesn't exist, should the status be changed from Pending to Error or Failed?  This would entail reverting to Pending once the error is corrected.

PR #170, opened 2019-09-24
renamed ob and obc phases to start with an upper case

### LINKED ISSUE #169: the phases of objectbucket and objectbucketclaim are not named according to conventions
phases names should begin with an upper case letter, similar to PVC phases names (`Pending`, `Bound`, etc. )

dannyzaken:
Fixes #169


copejon:
LGTM

PR #174, opened 2019-10-10
Fixed failing unit tests

copejon:
Updated stale unit tests.

### DIFF
--- a/pkg/provisioner/resourcehandlers_test.go
+++ b/pkg/provisioner/resourcehandlers_test.go
@@ -37,7 +37,7 @@ func TestNewCredentialsSecret(t *testing.T) {
 		authSecret   = "test-auth-secret"
 	)
 
-	var T = true
+	var isTrue = true
 	dummyLabels := map[string]string{"dummy": "label"}
 	testObjectMeta := metav1.ObjectMeta{
 		Name:       obcName,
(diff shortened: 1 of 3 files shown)


pkg/provisioner/resourcehandlers_test.go
```diff
@@ -37,11 +37,22 @@ func TestNewCredentialsSecret(t *testing.T) {
 		authSecret   = "test-auth-secret"
 	)
 
+	var T = true
```
jeffvance: isn't upper case `T` used to represent a generic type?

pkg/provisioner/resourcehandlers_test.go
```diff
@@ -137,17 +148,31 @@ func TestNewCredentialsSecret(t *testing.T) {
...
 
 	const (
+		obcName   = "test-obc"
 		host      = "http://www.test.com"
 		name      = "bucket-name"
 		port      = 11111
 		region    = "region"
 		subRegion = "sub-region"
 	)
+	var T = true
```
jeffvance: ditto

pkg/provisioner/resourcehandlers_test.go
copejon: ah, right. will fix

jeffvance:
LGTM

PR #177, opened 2019-12-03
Update to the k8s 1.16 client library

travisn:
In Rook I'm looking at updating to the K8s 1.16 client. Before Rook can make the change, the bucket provisioner must be updated. This updates to v1.16.3, including the corresponding changes to the generated code. 

In preliminary testing rook is building with these changes, but testing is still ongoing.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -10,16 +10,16 @@ required = [
 
 [[override]]
   name = "k8s.io/code-generator"
-  version = "kubernetes-1.14.1"
+  version = "kubernetes-1.16.3"
 
 [[constraint]]
   name = "k8s.io/api"
-  version = "kubernetes-1.14.1"
+  version = "kubernetes-1.16.3"
 
 [[constraint]]
   name = "k8s.io/apimachinery"
-  version = "kubernetes-1.14.1"
+  version = "kubernetes-1.16.3"
 
 [[constraint]]
   name = "k8s.io/client-go"
-  version = "kubernetes-1.14.1"
+  version = "kubernetes-1.16.3"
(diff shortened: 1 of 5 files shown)


guymguym:
Hey @travisn 

I thought the next steps should be:
1. Use go mod and get rid of go dep
2. Use operator-sdk v0.12.0 - https://github.com/operator-framework/operator-sdk/releases/tag/v0.12.0
3. Use controller-runtime v0.3.0 - https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.3.0

These dependencies mean to use kubernetes-1.15.4.

Is there a specific need to use 1.16 now rather than 1.15 which is supported by operator-fw?

travisn:
@guymguym The desire to use 1.16 was for this issue https://github.com/rook/rook/issues/4387 to start using the TopologySpreadConstraints on the pod spec. This feature is still only alpha in 1.16, but it would allow us to get an early start on this feature since it is critical for spreading the Ceph OSDs. 

Certainly we want to change to `go mod` and improve the dependency management and the controller runtime is certainly a critical dependency for this. I about have Rook compiling based on k8s 1.16.3 and v0.4 of the controller runtime. Was there a need for v0.3 of the controller runtime instead of v0.4? I'll continue with the prototype and see what I run into. If all goes well, perhaps we make this change first to unblock the topology spread constraints, then change dependency management as the next step.

guymguym:
controller-runtime v0.3 is just because this is the dependency of operator-sdk v0.12 which is the latest so far.
I am not sure what it would take to use controller-runtime v0.4 with operator-sdk...

But I think we should remove this dependency altogether.
I don't think lib-bucket-provisioner should restrict any project from using any runtime.
What if we keep this project to only contain the OB/OBC CRDs and the provisioner library itself would be abstracted out from the kubernetes API's completely so that rook can call it to reconcile an OBC struct. 
Without it I am afraid this will become a dependency hell.

travisn:
@guymguym Since lib-bucket-provisioner has a controller, how are you thinking it would be factored out? 

In any case, I'm going to hold off on this update to 1.16. The new feature it was needed for is still in alpha so it's not urgent to make the change now. Working through these dependencies first sounds better.

PR #178, opened 2019-12-06
Update to go modules

copejon:
This change will not affect consumers of the library.  Dependency management tools ignore the dependency management configs of imported packages.

Go modules significantly improve the developer experience by alleviating much of the dependency hell that arrises from working on multiple go projects at once.  The pain point is almost always related to the fact that before modules, the compiler expects to find one and only one version of a dependency in the GOPATH.   As the library deliberately does not manage `vendor/`, this means that developers are often having to cross the minefield that is the kubernetes versioned tool dependencies (client-go, apimachinery, etc) when they initialize their local env (`dep ensure`).

Secondly, go mods add a convenient method for substituting canonical import paths of go projects through the `replace` go.mod directive.   Replacements can be both working branches of forked repos or a relative path to the code on the host system.  This allows for code changes in the local copy of the library (on a forked repo and/or a working branch) to be built from directly by the dependee project.  This is far less messy than the current work flow, which is to copy/paste the entire library into the dependee's vendor directory in order to rebuild.

To enable this replace behavior, `go mod` expects a go.mod file to exist in the dependency.  It does not work otherwise.

Activating a replace from a repo fork and working branch can be done via the CLI.

`go mod edit --replace=github.com/kube-object-storate/lib-bucket-provisioner=github.com/myFork/lib-bucket-provisioner@myWorkingBranch`
Followed by
`go mod tidy`

or for a local path:
`go mod edit --replace=github.com/kube-object-storate/lib-bucket-provisioner=relative/PathTo/lib-bucket-provisioner`
Nothing else is required for local paths.


hack/README.md
```diff
@@ -75,45 +75,42 @@ Aggregates operations for execution by CI.  Right now this is `lint`, `test`, an
...
+
+    `$ go mod edit --replace=github.com/kube-object-bucket/lib-bucket-provisioner=$GOPATH/src/github.com/kube-object-storage/lib-bucket-provisioner`
+    
+1. build the s3 provisioner (from the s3-provisioner dir):
+   
+   `$ GOOS=linux GOARCH=amd64 go build -a -o ./bin/aws-s3-provisioner ./cmd/...`
+
+1. Start minikube:
+   
+   `$ minikube start --vm-driver=kvm2 --kubernetes-version=v1.14.0 --memory=5000 --cpus=4 && minikube update-context`
```
jeffvance: Probably the k8s version is not needed here?

hack/README.md
```diff
@@ -75,45 +75,42 @@ Aggregates operations for execution by CI.  Right now this is `lint`, `test`, an
...
+
+1. Build the container
```
jeffvance: maybe building the s3 container is more realistic way of testing, but the original hack approach was to simply run the s3 binary locally.

hack/README.md
copejon: The library is currently using v1.14.0, so it makes sense to avoid conflicts during runtime.

hack/README.md
copejon: True, though given that it's ultimately intended to run in k8s, it seemed like an antipattern to run it from the host machine.

jeffvance: fair

jeffvance:
LGTM. Can be merged.

PR #184, opened 2019-12-16
Fix infinite deletion loops and lock ups

copejon:
This PR address a couple OBC deletion related issues.

1. When an OBC is deleted before a finalizer can be placed on it, the IsNotFound error is now ignored.  The assumption is safe because, if the finalized had been set, then the OBC would not have vanished and thus the OBC would have been found.  Once the finalizer is place, it is only removed once all resources are cleaned up.  Thus, it is only possible for OBC to vanish in this small window if it was never processed to begin with. 

2. When resources cannot be found for deletion, IsNotFound errors are ignored.  In cases where a secret, cm, or ob are not found, continue on with deleting the rest of the resources and do not report the errors.

Signed-off-by: jcope <jcope@redhat.com>

### DIFF
--- a/pkg/provisioner/controller.go
+++ b/pkg/provisioner/controller.go
@@ -419,7 +419,7 @@ func (c *obcController) handleDeleteClaim(key string, obc *v1alpha1.ObjectBucket
 
 	log.Info("syncing obc deletion")
 
-	ob, cm, secret, errs := c.getResourcesForDeletionByKey(key)
+	ob, cm, secret, errs := c.getExistingResourcesFromKey(key)
 	if len(errs) > 0 {
 		return fmt.Errorf("error getting resources: %v", errs)
 	}
@@ -463,7 +463,7 @@ func (c *obcController) supportedProvisioner(provisioner string) bool {
 }
 
 // trim the errors resulting from objects not being found
-func (c *obcController) getResourcesForDeletionByKey(key string) (*v1alpha1.ObjectBucket, *corev1.ConfigMap, *corev1.Secret, []error) {
+func (c *obcController) getExistingResourcesFromKey(key string) (*v1alpha1.ObjectBucket, *corev1.ConfigMap, *corev1.Secret, []error) {
 	ob, cm, secret, errs := c.getResourcesFromKey(key)
 	for i := len(errs) - 1; i >= 0; i-- {
 		if errors.IsNotFound(errs[i]) {
@@ -479,6 +479,7 @@ func (c *obcController) getResourcesForDeletionByKey(key string) (*v1alpha1.Obje
 func (c *obcController) getResourcesFromKey(key string) (ob *v1alpha1.ObjectBucket, cm *corev1.ConfigMap, sec *corev1.Secret, errs []error) {
 
 	var err error
+	// The cap(errs) must be large enough to encapsulate errors returned by all 3 *ForClaimKey funcs
 	errs = make([]error, 0, 3)
 	groupErrors := func(err error) {
 		if err != nil {
--- a/pkg/provisioner/controller.go
+++ b/pkg/provisioner/controller.go
@@ -204,11 +204,24 @@ func (c *obcController) syncHandler(key string) error {
 
 	obc, err := claimForKey(key, c.libClientset)
 	if err != nil {
+		//      The OBC was deleted immediately after creation, before it could be processed by
+		//      handleProvisionClaim.  As a finalizer is immediately applied to the OBC before processing,
+		//      if it does not have a finalizer, it was not processed, and no artifacts were created.
+		//      Therefore, it is safe to assume nothing needs to be done.
 		if errors.IsNotFound(err) {
 			log.Info("OBC vanished, assuming it was deleted")
 			return nil
 		}
 		return fmt.Errorf("could not sync OBC %s: %v", key, err)
+	}
+
+	class, err := storageClassForClaim(c.clientset, obc)
+	if err != nil {
+		return err
+	}
+	if !c.supportedProvisioner(class.Provisioner) {
+		log.Info("unsupported provisioner", "got", class.Provisioner)
+		return nil
 	}
 
 	// ***********************
@@ -238,15 +251,6 @@ func (c *obcController) syncHandler(key string) error {
 		return fmt.Errorf("error updating OBC status: %s", err)
 	}
 
-	class, err := storageClassForClaim(c.clientset, obc)
-	if err != nil {
-		return err
-	}
-	if !c.supportedProvisioner(class.Provisioner) {
-		log.Info("unsupported provisioner", "got", class.Provisioner)
-		return nil
-	}
-
 	// By now, we should know that the OBC matches our provisioner, lacks an OB, and thus requires provisioning
 	err = c.handleProvisionClaim(key, obc, class)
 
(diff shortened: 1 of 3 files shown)


pkg/provisioner/controller.go
```diff
@@ -200,28 +200,23 @@ func (c *obcController) processNextItemInQueue() bool {
...
+	logD.Info("reconciling claim")
 
 	obc, err := claimForKey(key, c.libClientset)
 	if err != nil {
-		return fmt.Errorf("request key %q: %v", key, err)
-	}
-	if obc == nil {
-		return fmt.Errorf("unexpected nil obc for request key %q", key)
+		if errors.IsNotFound(err) {
+			log.Info("OBC vanished, assuming it was deleted")
```
jeffvance: how about adding the key value in this log msg?

pkg/provisioner/controller.go
```diff
@@ -200,28 +200,23 @@ func (c *obcController) processNextItemInQueue() bool {
...
+			return nil
+		}
+		return fmt.Errorf("could not sync OBC: %v", err)
```
jeffvance: and add the key here too unless you think knowing the obc ns/name won't help with debugging.

pkg/provisioner/controller.go
copejon: the obc key is encoded in the logger.  is there something more we should put in?

pkg/provisioner/controller.go
copejon: +1

pkg/provisioner/controller.go
```diff
@@ -261,12 +256,15 @@ func (c *obcController) syncHandler(key string) error {
...
-func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBucketClaim, class *storagev1.StorageClass) (err error) {
+func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBucketClaim, class *storagev1.StorageClass) error {
+
+	log.Info("syncing obc creation")
 
 	var (
 		ob        *v1alpha1.ObjectBucket
 		secret    *corev1.Secret
 		configMap *corev1.ConfigMap
+		err       error
```
jeffvance: Is there value defining `err` separately in the func body vs as a named return?

pkg/provisioner/controller.go
```diff
@@ -462,33 +462,38 @@ func (c *obcController) supportedProvisioner(provisioner string) bool {
...
+// Gathers resources by names derived from key.
+// Returns pointers to those resources if they exist, nil otherwise and an slice of errors who's
+// len() == n errors. If no errors occur, len() is 0.
+func (c *obcController) getResourcesFromKey(key string) (ob *v1alpha1.ObjectBucket, cm *corev1.ConfigMap, sec *corev1.Secret, errs []error) {
 
 	var err error
-	// return err if all resources were not retrieved, else no error
-	if obErr != nil && cmErr != nil && sErr != nil {
-		err = fmt.Errorf("could not get all needed resources: %v : %v : %v", obErr, cmErr, sErr)
+	errs = make([]error, 0, 3)
```
jeffvance: why 3 and why hard-coded?

pkg/provisioner/controller.go
```diff
@@ -462,33 +462,38 @@ func (c *obcController) supportedProvisioner(provisioner string) bool {
...
-	}
-	s, sErr := secretForClaimKey(key, c.clientset)
-	if errors.IsNotFound(sErr) {
-		log.Error(sErr, "secret not found")
-		sErr = nil
+// trim the errors resulting from objects not being found
+func (c *obcController) getResourcesForDeletionByKey(key string) (*v1alpha1.ObjectBucket, *corev1.ConfigMap, *corev1.Secret, []error) {
+	ob, cm, secret, errs := c.getResourcesFromKey(key)
+	for i := len(errs) - 1; i >= 0; i-- {
+		if errors.IsNotFound(errs[i]) {
```
jeffvance: What about passing an _ignoreNotFoundErr_ to `getResourcesFromKey` and then this func would not be needed?

pkg/provisioner/controller.go
```diff
@@ -283,7 +281,7 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	// It is assumed that if the get claim fails, no resources were generated to begin with.
 	defer func() {
 		if err != nil {
-			log.Error(err, "cleaning up reconcile artifacts")
+			log.Info("provisioning errored, cleaning up API resources")
```
jeffvance: `Info` instead of `Error` because Provision() errors are expected and handled?

pkg/provisioner/controller.go
jeffvance: No. I forgot about the key being in the logger.

pkg/provisioner/controller.go
copejon: I don't think it's necessarily idiomatic but my sense is that it hurts readability in large functions.  Particularly because we already have a block defining local variables.

pkg/provisioner/controller.go
copejon: it's specific to the amount of possible errors that can occur locally.  3 separate API calls in this case.

pkg/provisioner/controller.go
copejon: I'm channeling Dave Cheney here.  "Handle errors exactly once" - meaning either log them or return them.  I've spent a lot of time reading logs today and because it's both logged and returned, it's been difficult to tell from logs if it's occurring more than once.

pkg/provisioner/controller.go
jeffvance: Ack. Especially:
> Particularly because we already have a block defining local variables.

pkg/provisioner/controller.go
copejon: I see your point in that it could probably reduce lines of code and I think that overloads the intent of `getResourcesFromKey`, and perhaps it's reusability.   That said, naming the function specific to deletion also misstates what it does.  What do you think about delineating abstraction between "give me everything" and "give must just what exists"?

pkg/provisioner/controller.go
jeffvance: _"Handle errors exactly once"_ can be difficult sometimes but I agree it's a good philosophy,

pkg/provisioner/controller.go
jeffvance: I see now. I suppose there's no point in defining a global _number of k8s resources_ const... Maybe just add a comment here?

pkg/provisioner/controller.go
copejon: Agreed.  I think the line comes down to how ofter it appears in logs.  In this case, it was popping up 2 or 3 lines apart, several times before the reconcile iteration finished.

pkg/provisioner/controller.go
copejon: That's a good point, I'll drop one in here.

pkg/provisioner/controller.go
jeffvance: Yes, that abstraction works for me. +1

copejon:
Tested the changes against rook-ceph and aws-s3-provisioner by
1) running create/delete obc calls 1000 times back to back, over multiple iterations.

`for n in {1..1000}; do kubectl create -f <obc_file>; kubectl delete obc --all; done`

2) forcing a non-nil error return to the Provision() implementation in both provisioners, and rerunning 1). 

Prior to this fix, I would expect 1 of 2 cases to occur:
a) the OBC would vanish but the library would have continued to try to process the claim, resulting in infinite retry looping
b) the CLI would hang indefinitely as a result of the library not properly responding to the deletionTimestamp.

Both of these cases are resolved.  
Deleting the OBC before provisioning is complete triggers a follow-on Delete() sync.  
Deleting the OBC during an error back off loop now responds appropriately by cleaning up library artifacts and ending the loop.

pkg/provisioner/controller.go
jeffvance: Is the note comment above still accurate?

pkg/provisioner/controller.go
```diff
@@ -200,28 +200,23 @@ func (c *obcController) processNextItemInQueue() bool {
...
 
 	obc, err := claimForKey(key, c.libClientset)
 	if err != nil {
-		return fmt.Errorf("request key %q: %v", key, err)
-	}
-	if obc == nil {
-		return fmt.Errorf("unexpected nil obc for request key %q", key)
+		if errors.IsNotFound(err) {
+			log.Info("OBC vanished, assuming it was deleted")
+			return nil
```
jeffvance: In this scenario will the lib clean up the OB, secret, and CM if they still exist?

pkg/provisioner/controller.go
```diff
@@ -200,28 +200,23 @@ func (c *obcController) processNextItemInQueue() bool {
...
+	// Delete or Revoke Bucket
+	// ***********************
+	if obc.ObjectMeta.DeletionTimestamp != nil {
 		log.Info("OBC deleted, proceeding with cleanup")
-		err = c.handleDeleteClaim(key, obc)
-		if err != nil {
-			log.Error(err, "error cleaning up OBC", "name", key)
-		}
-		return err
+		return c.handleDeleteClaim(key, obc)
```
jeffvance: In the `updateObjectBucketClaimPhase` call below _(sorry I don't know how to add a comment to line that wasn't changed)_ why do we update the phase before we know if the provisioner name matches? It's possible, isn't it, that this obc is for a different provisioner and we've just updated it?

pkg/provisioner/controller.go
```diff
@@ -283,7 +281,7 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	// It is assumed that if the get claim fails, no resources were generated to begin with.
 	defer func() {
 		if err != nil {
-			log.Error(err, "cleaning up reconcile artifacts")
+			log.Info("provisioning errored, cleaning up API resources")
 			if !pErr.IsBucketExists(err) && ob != nil && isDynamicProvisioning {
```
jeffvance: maybe add comment that the ob is non-nil after call to `Provision` or `Revoke` and *before* any k8s resources have been created. I forgot this and spent a little time thinking about error cases where secret or cm have been created before the ob...

pkg/provisioner/controller.go
copejon: Right.  If the OBC vanished, then it must not have had a finalizer.  And if there was no finalizer, then never made it to the beginning of the sync, and so no resources would have been created.

+1 for comment, will add

pkg/provisioner/controller.go
copejon: +100, no idea if or when that moved. will fix

pkg/provisioner/controller.go
copejon: +1

pkg/provisioner/controller.go
copejon: fixed

pkg/provisioner/controller.go
copejon: fixed

jeffvance:
LGTM

Issue #189, opened 2020-05-21
Do not block OBC request handling on OBC deletion


thotz:
Please refer following rook [issue](https://github.com/rook/rook/issues/4907)

thotz:
@jeffvance @copejon I know this library is planning to depreciated. Do I need on [cosi](https://github.com/kubernetes/enhancements/pull/1383) and work on it later instead of fixing it now

copejon:
COSI is still a long way off in terms of its first alpha release.  For now, I would suggest running concurrent threads of workers.  This change would mean that the [Start() function](https://github.com/kube-object-storage/lib-bucket-provisioner/blob/02cba53742aed9a57c6541bb8dbf97346bc92835/pkg/provisioner/controller.go#L118) in the library could be altered to be optionally run in >1 go routines as it's done in the [sample controller](https://github.com/kubernetes/sample-controller/blob/a1a67935a65841952d990076df30b6649f25bf53/controller.go#L166).

So the line in the library 
```go
go wait.Until(c.runWorker, time.Second, stopCh)
```

would be wrapped with a loop and a parameter added to set threadiness, as it is in the sample controller:
```go
func Run(threadiness int, stopCh <-chan struct{}) {
...
    for i := 0; i < threadiness; i++ {
       	    go wait.Until(c.runWorker, time.Second, stopCh)
    }
...
}
```

The default should be 1.  When rook starts the library controller, it can pass in a higher number of threads.

thotz:
> COSI is still a long way off in terms of its first alpha release.  For now, I would suggest running concurrent threads of workers.  This change would mean that the [Start() function](https://github.com/kube-object-storage/lib-bucket-provisioner/blob/02cba53742aed9a57c6541bb8dbf97346bc92835/pkg/provisioner/controller.go#L118) in the library could be altered to be optionally run in >1 go routines as it's done in the [sample controller](https://github.com/kubernetes/sample-controller/blob/a1a67935a65841952d990076df30b6649f25bf53/controller.go#L166).
> 
> So the line in the library 
> ```go
> go wait.Until(c.runWorker, time.Second, stopCh)
> ```
> 
> would be wrapped with a loop and a parameter added to set threadiness, as it is in the sample controller:
> ```go
> func Run(threadiness int, stopCh <-chan struct{}) {
> ...
>     for i := 0; i < threadiness; i++ {
>        	    go wait.Until(c.runWorker, time.Second, stopCh)
>     }
[...]

PR #192, opened 2020-06-05
support multithreading for provisioner #189

thotz:


thotz:
@copejon @jeffvance please review

copejon:
Good looks good.  Has it been tested with rook-ceph?

thotz:
> Good looks good.  Has it been tested with rook-ceph?
> 

 @copejon  Yes with 8 threads

pkg/provisioner/controller.go
```diff
@@ -115,8 +117,16 @@ func (c *obcController) Start(stopCh <-chan struct{}) error {
 	if !cache.WaitForCacheSync(stopCh, c.obcHasSynced, c.obHasSynced) {
 		return fmt.Errorf("failed to waith for caches to sync ")
 	}
-	go wait.Until(c.runWorker, time.Second, stopCh)
-
+	var count int
+	threadiness, set := os.LookupEnv("LIB_BUCKET_PROVISIONER_THREADS")
```
jeffvance, line 121: It is more typical to see ENV variables collected in _main_ and then passed as args to other functions. If we did this then every provisioner that wanted >1 thread would have to either get the LIB_BUCKET_PROVISIONER_THREADS env variable or hard-code their value (with 1 being the lib's default). @copejon @guymguym any thoughts on this?

pkg/provisioner/controller.go
```diff
@@ -115,8 +117,16 @@ func (c *obcController) Start(stopCh <-chan struct{}) error {
 	if !cache.WaitForCacheSync(stopCh, c.obcHasSynced, c.obHasSynced) {
 		return fmt.Errorf("failed to waith for caches to sync ")
 	}
-	go wait.Until(c.runWorker, time.Second, stopCh)
-
+	var count int
```
jeffvance, line 120: More _go_ like as follows:
```
count := 1
threadiness, set := os.LookupEnv("LIB_BUCKET_PROVISIONER_THREADS")
if set {
	threadiness, set := os.LookupEnv("LIB_BUCKET_PROVISIONER_THREADS")
}
```

thotz: Done @jeffvance

jeffvance:
LGTM

@copejon @guymguym still interested in your thought on this [comment](https://github.com/kube-object-storage/lib-bucket-provisioner/pull/192#discussion_r436020112)

pkg/provisioner/controller.go
copejon: The library has now `main` function, but rather just the Run() method.  

> If we did this then every provisioner that wanted >1 thread would have to either get the LIB_BUCKET_PROVISIONER_THREADS

Not sure I follow.  The env var is defined and accessed by the library.  There's no need for dependees to also have to define/get it.  Rather, it's just something that would be optionally set in the runtime environment.  If not set, defaults to 1.

jeffvance: I was wondering (out loud) if the lib should _not_ get the LIB_BUCKET_PROVISIONER_THREADS env variable. Instead, each provisioner passes to the lib the # of threads they want with the lib default remaining 1. Then each provisioner decides how to determine their own # of threads.

copejon: If the library defines an env var, but then requires that dependees get the value and pass it to the constructor rather than checking it itself, then the env var doesn't actually have any purpose.  Since the value is passed through the constructor function, where it's set is then completely arbitrary.

This also defeats the purpose of using an env var, which was to not make changes to the library interface.  So if we're defining an env var (which I still think we should), then the library should access it directly.

jeffvance: 1) I was thinking the lib would _not_ define an env var,
2) `defeats the purpose of using an env var, which was to not make changes to the library interface` I completely agree and retract my comment.

jeffvance:
@thotz Ok to merge after squashing commits.

PR #197, opened 2020-11-13
dep: bump to k8s 1.19

leseb:
Consumers of the library like Rook are bumping their deps version to
Kubernetes 1.19. So the lib must be compatible with it.

I have to admit that this is not the ideal way to pass context,
typically it should come from the caller. I had investigated that route
but the required changes are enormous. Also, a lot of the build codegen
code is broken. Due the deprecation state of the lib is abandoned that
option and thus want with this "good enough" approach.


leseb:
@guymguym @dannyzaken @jeffvance @copejon @jackyalbo PATL.

PR #198, opened 2020-11-19
fix errors during provisioning

BlaineEXE:
Multiple fixes for issue reported in https://github.com/rook/rook/issues/6650 as well as issues found during investigation of aforementioned bug.

When provisioning fails and the deferred cleanup function is called, the
controller must take care to ensure the object bucket resource being
cleaned up (deleted) has the necessary fields filled in for delete calls
to be successful. Fill in basic object bucket info as soon as possible
after Privision() or Grant() returns, and in the event of an error
during one of these calls, make sure the deferred cleanup function also
sets the information it can before calling Delete().

The code generation script was not up to date for the k8s v1.19 API.
Update code gen to work again, and update the generated code.
This fixes a bug where provisioning completes and should not start
again, but the controller allows provisioning to complete forever.

Tagging: @leseb @travisn @thotz 
They will be interested to comment on this PR.

Related: https://github.com/rook/rook/pull/6672


BlaineEXE:
@jeffvance  @copejon  @thotz

pkg/provisioner/controller.go
```diff
@@ -351,7 +365,15 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	if err != nil {
 		return fmt.Errorf("error %s bucket: %v", verb, err)
 	} else if ob == (&v1alpha1.ObjectBucket{}) {
-		return fmt.Errorf("provisioner returned nil/empty object bucket")
+		return fmt.Errorf("provisioner returned empty object bucket")
+	}
+
+	// Fill in known information in the object bucket struct as soon as possible
+	setBasicOBInfo()
```
jeffvance, line 372: If the provisioner were to set ob.Spec.ReclaimPolicy, say based on some flag or config, we will overwrite that now.

pkg/provisioner/controller.go
```diff
@@ -337,6 +318,39 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+		setObjectBucketName(ob, key)
+		ob.Spec.StorageClassName = obc.Spec.StorageClassName
+		ob.Spec.ReclaimPolicy = options.ReclaimPolicy
+		ob.SetLabels(c.provisionerLabels)
+	}
+
+	// Should an error be returned, attempt to clean up the object store and API servers by
+	// calling the appropriate provisioner method.  In cases where Provision() or Revoke()
+	// return an err, it's likely that the ob == nil, hindering cleanup.
+	defer func() {
```
jeffvance, line 332: can this block be moved even further down in the func, say right before the secret is created?

pkg/provisioner/controller.go
```diff
@@ -406,7 +421,8 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 
 	// update OBC
 	obc.Spec.ObjectBucketName = ob.Name
-	obc.Spec.BucketName = bucketName
+	// Do not set obc.Spec.BucketName here: (1) if user has it set, it should be as they set it
```
thotz, line 424: Good Points

thotz (approved):
LGTM

pkg/provisioner/controller.go
BlaineEXE: I didn't modify any of the functional code that set the spec, so this would have been the case all along. I will add another commit to fix that small issue here also.

pkg/provisioner/controller.go
BlaineEXE: I think it **can**, but this will mean that if the implementation of `Provision` or `Grant` doesn't clean up fully, then there will not be a later `Delete`/`Revoke` call made. I would err on the side of keeping this where it is to allow for more "slop" in the `Provision`/`Grant` implementation, but I am open to moving this to just before secret creation and updating the `Provision` and `Grant` docs to mention that if there are any errors during `Provision`/`Grant`, the implementation **must** fully clean up all of its resources.

In responding here, my thoughts do raise another consideration that is out of scope for this PR. If I imagine a case where the operator is `kill -9`ed or there is a full crash of the operator, it's possible that the object bucket provisioning will have created a bucket and user for the bucket that get orphaned (they aren't recorded as created but also aren't cleaned up after creation). I wonder if it would be good to have a cleanup cycle on operator start where Rook and/or the provisioner would be able to find orphaned users/buckets/ObjectBuckets to delete them.

jeffvance: In the past we thought about a cleanup cycle. I think it's probably a good idea.

pkg/provisioner/controller.go
BlaineEXE: Looking into this further, I'm not sure this is an issue. Currently the reclaim policy is set here in the lib controller from the storage class. If the provisioner implementation sets its own reclaim policy that is different from the storage class, that to me seems like a bug. But perhaps there is a case I can't think of where Rook or another provisioner implementation would want to take override the reclaim policy. I think such a case might confuse users, however.

jeffvance: I think it would be unusual for the provisioner to modify the OB, but it could and there might be a good reason to do so in certain cases. If  the lib wants to prohibit this then maybe it should call `setBasicOBInfo` after the provisioner returns?

pkg/provisioner/controller.go
```diff
@@ -337,6 +318,43 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
 
+	// Fill in basic necessary object bucket info
+	setBasicOBInfo := func() {
+		setObjectBucketName(ob, key)
+		ob.Spec.StorageClassName = obc.Spec.StorageClassName
+		if ob.Spec.ReclaimPolicy == nil || (ob.Spec.ReclaimPolicy != nil && *ob.Spec.ReclaimPolicy == corev1.PersistentVolumeReclaimPolicy("")) {
+			// Do not blindly overwrite the reclaim policy. The provisioner might have reason to
+			// specify a reclaim policy that is  different from the storage class.
+			ob.Spec.ReclaimPolicy = options.ReclaimPolicy
+		}
```
BlaineEXE, line 329: @jeffvance what do you think about doing a check to set (or not set) the reclaim policy only if the Provision implementation hasn't already set it?

Doing it this way means that in the case where Provision fails but still returns an OB, the OB will still get the default reclaim policy from the storage class before cleanup is called on it. The only possible hangup being that if the Provision implementation normally wants to override the cleanup policy, if it does not set its desired cleanup policy before the failure, this lib's controller will set the value. I don't foresee this being much of an issue. If the Provision implementation wants to specify a cleanup policy but doesn't do so in all cases including error cases where it still returns an OB, I view that as a bug in the Provision implementation.

pkg/provisioner/api/provisioner.go
```diff
@@ -26,7 +26,15 @@ import (
...
-	// Provision should be implemented to handle bucket creation
+	// Provision should be implemented to handle creation of object storage buckets.
+	// Provision should NOT create the ObjectBucket resource. ObjectBucket resource creation is done
+	// by this library's controller.
+	// The Provision implementation must return an ObjectBucket struct with at least the Connection
+	// spec filled in. All other ObjectBucket details will be filled in by this library's controller
+	// before the ObjectBucket resource is created.
+	// The Provision implementation may opt to specify the ObjectBucket spec's ReclaimPolicy in
+	// cases where the provisioner wishes to set a different value from the one specified in the
+	// ObjectBucketClaim's StorageClass.
```
BlaineEXE, line 37: @jeffvance I have updated the API documentation for `Provision` to clarify more details about how it should be implemented. Let me know if I should also (or alternately) update info in other places.

pkg/provisioner/api/provisioner.go
```diff
@@ -26,7 +26,20 @@ import (
...
+	// The Provision implementation should clean up its own resources and return a nil ObjectBucket
+	// struct when returning an error. However, if Provision returns an error, it may optionally
+	// return a non-nil, non-empty ObjectBucket struct. In this case, this library's controller will
+	// specify the ObjectBucket's Name, StorageClassName, ReclaimPolicy, and Labels and then call
+	// Delete() with the ObjectBucket to clean up resources.
 	Provision(options *BucketOptions) (*v1alpha1.ObjectBucket, error)
```
BlaineEXE, line 43: @jeffvance I have updated the API documentation for `Provision` to clarify more details about how it should be implemented. Let me know if I should also (or alternately) update info in other places.

pkg/provisioner/controller.go
jeffvance: I accidentally clicked `Resolve` and can't figure out how to undo this.

pkg/provisioner/controller.go
```diff
@@ -337,6 +318,43 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		Parameters:        class.Parameters,
 	}
 
+	// Fill in basic necessary object bucket info
+	setBasicOBInfo := func() {
+		setObjectBucketName(ob, key)
+		ob.Spec.StorageClassName = obc.Spec.StorageClassName
+		if ob.Spec.ReclaimPolicy == nil || (ob.Spec.ReclaimPolicy != nil && *ob.Spec.ReclaimPolicy == corev1.PersistentVolumeReclaimPolicy("")) {
```
jeffvance, line 325: nit: `if` can be rewritten as:
```
if ob.Spec.ReclaimPolicy == nil || *ob.Spec.ReclaimPolicy == corev1.PersistentVolumeReclaimPolicy("") {
```

pkg/provisioner/controller.go
jeffvance: @BlaineEXE I thinks this is an improvement to handle a low probability use-case. Note that today the _reclaimPolicy_ does not influence lib cleanup, but it might in the future, and if that change were to be implemented then we'd want the `setBasicOBInfo` func to also check `err`.

jeffvance:
lgtm
Thanks @BlaineEXE

BlaineEXE:
Seems like this is approved. How can I get it merged now?

BlaineEXE:
@jeffvance are you able to do the merge?

Also, since Rook makes use of this library still, maybe we can chat about what adding a maintainer to this library would look like so we don't have to bug you when we need minor changes.

jeffvance:
> maybe we can chat about what adding a maintainer to this library 

Agree. Maybe you and/or @thotz  and/or Travis?

thotz:
> > maybe we can chat about what adding a maintainer to this library
> 
> Agree. Maybe you and/or @thotz and/or Travis?

Even I agree that it better to have someone from Rook project as maintainer here

PR #200, opened 2020-12-04
Allow provisioning to be idempotent

BlaineEXE:
If Provisioner implementers fail to include the `objectBucketName`
parameter as part of the validated ObjectBucketClaim CRD, this library
will not be able to determine if a claim is difinitively bound. In this
case, the library will continue to try to provision claims that have
already been provisioned. This was the case in Rook with the bug below.

https://github.com/rook/rook/issues/6650

If an implementing Provisioner fails to include "objectBucketName" in
the CRDs needed for object bucket claims, Kubernetes may drop the
unknown "objectBucketName" parameter that this library uses to determine
if an OBC is bound to an OB. Allow this library to recover from this
case or other degraded cases that are similar by checking whether the
Object Bucket that would be created for the claim already exists.

If the OB exists during provisioning and has an owner reference to the OBC,
then do not continue provisioning. Instead, attempt to update the OBC to
refer to the OB via "objectBucketName", and update the phase to "Bound".

Another unlikely but possible case where this degraded case might be
encountered if the Provisioner operator crashes after updating the OB but
before the OBC can be updated. The changes here will also help recover
from this edge case.

Also:

When provisioning an OBC, the library can fail on creation of secret or
configmap in the case where the operator crashes after secret and/or
configmap are created but before provisioning is finished. Allow the
provision code to find existing secrets and/or configmaps and continue
with provisioning if they exist rather than erroring out to infinitely
fail provisioning in a loop.

@thotz @leseb @travisn @guymguym @jeffvance @copejon

### DIFF
--- a/pkg/provisioner/controller.go
+++ b/pkg/provisioner/controller.go
@@ -302,13 +302,20 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		return fmt.Errorf("bucket name missing")
 	}
 
-	// Re-Get the claim in order to shorten the race condition where the claim was deleted after provisioning started
-	obc, err = claimForKey(key, c.libClientset)
-	if err != nil {
-		if errors.IsNotFound(err) {
-			return fmt.Errorf("OBC was lost before we could provision: %v", err)
+	// In the case where a bucket name is being generated, generate the name and store it in the OBC
+	// spec before doing any Provisioning so that any crashes encountered in this code will not
+	// result in multiple buckets being generated for the same OBC. bucketName takes precedence over
+	// generateBucketName if both are present.
+	if obc.Spec.BucketName == "" {
+		obc.Spec.BucketName = bucketName
+		obc, err = updateClaim(
+			c.libClientset,
+			obc,
+			defaultRetryBaseInterval,
+			defaultRetryTimeout)
+		if err != nil {
+			return fmt.Errorf("error updating OBC %q with bucket name: %v", key, err)
 		}
-		return err
 	}
 
 	updateOnSuccess := func() error {
--- a/pkg/provisioner/controller.go
+++ b/pkg/provisioner/controller.go
@@ -311,6 +311,58 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		return err
 	}
 
+	updateOnSuccess := func() error {
+		// In the case where the OB has been created and the operator crashed immediately after
+		// (before the OB phase could be updated), we should update the OB phase in addition to
+		// updating the OBC reference to the OB and the OBC's phase.
+		ob, err = updateObjectBucketPhase(
+			c.libClientset,
+			ob,
+			v1alpha1.ObjectBucketStatusPhaseBound,
+			defaultRetryBaseInterval,
+			defaultRetryTimeout)
+		if err != nil {
+			return fmt.Errorf("error updating OB %q's status to %q: %v", ob.Name, v1alpha1.ObjectBucketStatusPhaseBound, err)
+		}
+		// update OBC
+		obc.Spec.ObjectBucketName = ob.Name
+		obc.Spec.BucketName = bucketName
+		obc, err = updateClaim(
+			c.libClientset,
+			obc,
+			defaultRetryBaseInterval,
+			defaultRetryTimeout)
+		if err != nil {
+			return fmt.Errorf("error updating OBC: %v", err)
+		}
+		obc, err = updateObjectBucketClaimPhase(
+			c.libClientset,
+			obc,
+			v1alpha1.ObjectBucketClaimStatusPhaseBound,
+			defaultRetryBaseInterval,
+			defaultRetryTimeout)
+		if err != nil {
+			return fmt.Errorf("error updating OBC %q's status to: %v", v1alpha1.ObjectBucketClaimStatusPhaseBound, err)
+		}
+		return nil
+	}
+
+	// recover from degraded state where provisioning is happening again on a completed obc
+	ob, err = getObjectBucketFromClaimKey(key, obc, c.libClientset, defaultRetryBaseInterval, defaultRetryTimeout)
+	if err != nil {
+		return fmt.Errorf("error getting OB for OBC %q: %v", key, err)
+	}
+	if ob != nil {
+		err = updateOnSuccess()
+		if err != nil {
+			return fmt.Errorf("error marking OB %q as bound to OBC %q: %v", ob.Name, key, err)
+		}
+		// Do not do any more provisioning; the bucket is already provisioned and merely
+		// needed its info re-updated to reference the claim and vice versa
+		logD.Info("(re-)bound OB to OBC successfully", "OB", ob.Name)
+		return nil
+	}
+
 	options := &api.BucketOptions{
 		ReclaimPolicy:     class.ReclaimPolicy,
 		BucketName:        bucketName,
@@ -380,7 +432,18 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		return fmt.Errorf("error getting reference to OBC: %v", err)
 	}
 
-	// create Secret and ConfigMap
+	// Create auth Secret and bucket info ConfigMap
+	// If the operator crashes before the OB is created, provisioning will happen again including a
+	// call to Provision()/Grant(), and a Secret and/or ConfigMap might exist from the previous
+	// provisioning attempt. Provision() isn't guaranteed to give the same credentials or bucket
+	// each time, so if the Secret/ConfigMap already exists, delete the old one and re-create it
+	// with the latest info.
+
+	err = deleteExistingSecretAndConfigMapIfExist(obc, c.clientset, defaultRetryBaseInterval, defaultRetryTimeout)
+	if err != nil {
+		return fmt.Errorf("failed to delete existing Secret and/or ConfigMap for OBC %q: %v", obc.Namespace+"/"+obc.Name, err)
+	}
+
 	secret, err = createSecret(
 		obc,
 		ob.Spec.Authentication,
@@ -391,6 +454,7 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	if err != nil {
 		return fmt.Errorf("error creating secret for OBC: %v", err)
 	}
+
 	configMap, err = createConfigMap(
 		obc,
 		ob.Spec.Endpoint,
@@ -413,36 +477,10 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	if err != nil {
 		return fmt.Errorf("error creating OB %q: %v", ob.Name, err)
 	}
-	ob, err = updateObjectBucketPhase(
-		c.libClientset,
-		ob,
-		v1alpha1.ObjectBucketStatusPhaseBound,
-		defaultRetryBaseInterval,
-		defaultRetryTimeout)
-	if err != nil {
-		return fmt.Errorf("error updating OB %q's status to %q: %v", ob.Name, v1alpha1.ObjectBucketStatusPhaseBound, err)
-	}
 
-	// update OBC
-	obc.Spec.ObjectBucketName = ob.Name
-	// Do not set obc.Spec.BucketName here: (1) if user has it set, it should be as they set it
-	// (2) if generateBucketName is set, setting this will cause errors if provisioning runs again
-	obc, err = updateClaim(
-		c.libClientset,
-		obc,
-		defaultRetryBaseInterval,
-		defaultRetryTimeout)
+	err = updateOnSuccess()
 	if err != nil {
-		return fmt.Errorf("error updating OBC: %v", err)
-	}
-	obc, err = updateObjectBucketClaimPhase(
-		c.libClientset,
-		obc,
-		v1alpha1.ObjectBucketClaimStatusPhaseBound,
-		defaultRetryBaseInterval,
-		defaultRetryTimeout)
-	if err != nil {
-		return fmt.Errorf("error updating OBC %q's status to: %v", v1alpha1.ObjectBucketClaimStatusPhaseBound, err)
+		return fmt.Errorf("error marking new OB %q as bound to OBC %q: %v", ob.Name, key, err)
 	}
 
 	log.Info("provisioning succeeded")
--- a/pkg/provisioner/resourcehandlers.go
+++ b/pkg/provisioner/resourcehandlers.go
@@ -66,7 +66,7 @@ func newBucketConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, 
 
 	return &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:       obc.Name,
+			Name:       composeConfigMapName(obc),
 			Namespace:  obc.Namespace,
 			Finalizers: []string{finalizer},
 			Labels:     labels,
@@ -99,7 +99,7 @@ func newCredentialsSecret(obc *v1alpha1.ObjectBucketClaim, auth *v1alpha1.Authen
 
 	secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:       obc.Name,
+			Name:       composeSecretName(obc),
 			Namespace:  obc.Namespace,
 			Finalizers: []string{finalizer},
 			Labels:     labels,
@@ -129,6 +129,55 @@ func createObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface, retryI
 		return (err == nil), err
 	})
 	return
+}
+
+func updateObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface, retryInterval, retryTimeout time.Duration) (result *v1alpha1.ObjectBucket, err error) {
+	logD.Info("updating ObjectBucket", "name", ob.Name)
+	result, err = c.ObjectbucketV1alpha1().ObjectBuckets().Update(context.TODO(), ob, metav1.UpdateOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return result, err
+}
+
+func getObjectBucketFromClaimKey(key string, obc *v1alpha1.ObjectBucketClaim, c versioned.Interface, retryInterval, retryTimeout time.Duration) (ob *v1alpha1.ObjectBucket, err error) {
+	obName, err := objectBucketNameFromClaimKey(key)
+	if err != nil {
+		return nil, err
+	}
+
+	logD.Info("seeing if OB for OBC exists", "checking for OB name", obName)
+	ob, err = c.ObjectbucketV1alpha1().ObjectBuckets().Get(context.TODO(), obName, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Not found is a valid exit here which returns no error and a nil result
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	if bucketIsOwnedByClaim(obc, ob) {
+		logD.Info("found OB belonging to this OBC", "OB", ob.Name)
+		return ob, err
+	}
+
+	emptyRef := corev1.ObjectReference{}
+	if ob.Spec.ClaimRef == nil || *ob.Spec.ClaimRef == emptyRef {
+		// If the CRD validation for the OB improperly specifies the object reference field, it will
+		// be missing. Do our best to recover from this case gracefully.
+		logD.Info("found OB matching OBC, but the OB ClaimRef is empty. binding OB to OBC", "OB", ob.Name)
+		ob.Spec.ClaimRef, err = claimRefForKey(key, c)
+		if err != nil {
+			return nil, fmt.Errorf("error generating OB %q's reference to OBC: %v", ob.Name, err)
+		}
+		ob, err = updateObjectBucket(ob, c, defaultRetryBaseInterval, defaultRetryTimeout)
+		if err != nil {
+			return nil, fmt.Errorf("error updating OB %q with assumed claim ref: %v", ob.Name, err)
+		}
+		return ob, err
+	}
+
+	return nil, fmt.Errorf("found OB %q not owned by OBC %q, likely an artifact from a previous OBC that did not get cleaned up, user must delete the OB in order to allow OBC provisioning to continue", ob.Name, key)
 }
 
 func createSecret(obc *v1alpha1.ObjectBucketClaim, auth *v1alpha1.Authentication, labels map[string]string, c kubernetes.Interface, retryInterval, retryTimeout time.Duration) (*corev1.Secret, error) {
@@ -174,6 +223,137 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, lab
 		return true, nil
 	})
 	return configMap, err
+}
+
+func deleteExistingSecretAndConfigMapIfExist(obc *v1alpha1.ObjectBucketClaim, c kubernetes.Interface, retryInterval, retryTimeout time.Duration) error {
+
+	secret, err := getExistingSecret(obc, c, defaultRetryBaseInterval, defaultRetryTimeout)
+	if err != nil {
+		return fmt.Errorf("error checking for existing OBC credentials Secret: %v", err)
+	}
+	if secret != nil {
+		if !objectIsOwnedByClaim(obc, secret.OwnerReferences) {
+			return fmt.Errorf("found OBC credentials Secret %q not owned by OBC, assuming this is a user-created resource, user must delete the Secret in order to allow OBC provisioning to continue", secret.Name)
+		}
+		logD.Info("found OBC credentials Secret owned by OBC, deleting and re-creating")
+		err = deleteSecretAndWait(obc, c, defaultRetryBaseInterval, defaultRetryTimeout)
+		if err != nil {
+			return fmt.Errorf("failed to delete existing OBC credentials Secret in order to re-create: %v", err)
+		}
+	}
+
+	configMap, err := getExistingConfigMap(obc, c, defaultRetryBaseInterval, defaultRetryTimeout)
+	if err != nil {
+		return fmt.Errorf("error checking for existing OBC bucket info ConfigMap: %v", err)
+	}
+	if configMap != nil {
+		if !objectIsOwnedByClaim(obc, configMap.OwnerReferences) {
+			return fmt.Errorf("found OBC bucket info ConfigMap %q not owned by OBC, assuming this is a user-created resource, user must delete the ConfigMap in order to allow OBC provisioning to continue", configMap.Name)
+		}
+		logD.Info("found OBC bucket info ConfigMap owned by OBC, deleting and re-creating")
+		err = deleteConfigMapAndWait(obc, c, defaultRetryBaseInterval, defaultRetryTimeout)
+		if err != nil {
+			return fmt.Errorf("failed to delete existing OBC bucket info ConfigMap in order to re-create: %v", err)
+		}
+	}
+
+	return nil
+}
+
+func getExistingSecret(obc *v1alpha1.ObjectBucketClaim, c kubernetes.Interface, retryInterval, retryTimeout time.Duration) (secret *corev1.Secret, err error) {
+	secret, err = c.CoreV1().Secrets(obc.Namespace).Get(context.TODO(), composeSecretName(obc), metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Not found is valid here which returns no error and a nil secret
+			return nil, nil
+		}
+		return nil, err
+	}
+	return secret, err
+}
+
+func getExistingConfigMap(obc *v1alpha1.ObjectBucketClaim, c kubernetes.Interface, retryInterval, retryTimeout time.Duration) (configMap *corev1.ConfigMap, err error) {
+	configMap, err = c.CoreV1().ConfigMaps(obc.Namespace).Get(context.TODO(), composeConfigMapName(obc), metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Not found is valid here which returns no error and a nil configMap
+			return nil, nil
+		}
+		return nil, err
+	}
+	return configMap, err
+}
+
+func deleteSecretAndWait(obc *v1alpha1.ObjectBucketClaim, c kubernetes.Interface, retryInterval, retryTimeout time.Duration) error {
+	name := composeSecretName(obc)
+	logD.Info("deleting secret", "secret name", name)
+
+	secret, err := c.CoreV1().Secrets(obc.Namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Already deleted
+			return nil
+		}
+		return fmt.Errorf("failed to get secret in order to delete it: %v", err)
+	}
+	err = releaseSecret(secret, c)
+	if err != nil {
+		return fmt.Errorf("failed to release secret for deletion: %v", err)
+	}
+	err = c.CoreV1().Secrets(obc.Namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
+	if err != nil {
+		return err
+	}
+	err = wait.PollImmediate(retryInterval, retryTimeout, func() (done bool, err error) {
+		_, err = c.CoreV1().Secrets(obc.Namespace).Get(context.TODO(), name, metav1.GetOptions{})
+		if err != nil {
+			if errors.IsNotFound(err) {
+				// Is Deleted
+				err = nil
+				return true, nil
+			}
+			// for other errors, give the opportunity to retry until timeout
+			return false, nil
+		}
+		return false, nil // still exists
+	})
+	return err
+}
+
+func deleteConfigMapAndWait(obc *v1alpha1.ObjectBucketClaim, c kubernetes.Interface, retryInterval, retryTimeout time.Duration) error {
+	name := composeConfigMapName(obc)
+	logD.Info("deleting configmap", "configmap name", name)
+
+	configMap, err := c.CoreV1().ConfigMaps(obc.Namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Already deleted
+			return nil
+		}
+		return fmt.Errorf("failed to get ConfigMap in order to delete it: %v", err)
+	}
+	err = releaseConfigMap(configMap, c)
+	if err != nil {
+		return fmt.Errorf("failed to release ConfigMap for deletion: %v", err)
+	}
+	err = c.CoreV1().ConfigMaps(obc.Namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
+	if err != nil {
+		return err
+	}
+	err = wait.PollImmediate(retryInterval, retryTimeout, func() (done bool, err error) {
+		_, err = c.CoreV1().ConfigMaps(obc.Namespace).Get(context.TODO(), name, metav1.GetOptions{})
+		if err != nil {
+			if errors.IsNotFound(err) {
+				// Is Deleted
+				err = nil
+				return true, nil
+			}
+			// for other errors, give the opportunity to retry until timeout
+			return false, nil
+		}
+		return false, nil // still exists
+	})
+	return err
 }
 
 // Only the finalizer needs to be removed. The CM will be garbage collected since its
--- a/pkg/provisioner/helpers.go
+++ b/pkg/provisioner/helpers.go
@@ -60,9 +60,35 @@ func makeOwnerReference(claim *v1alpha1.ObjectBucketClaim) metav1.OwnerReference
 	}
 }
 
+// return true if owner refs report being owned by the obc, false otherwise
+func objectIsOwnedByClaim(obc *v1alpha1.ObjectBucketClaim, ownerRefs []metav1.OwnerReference) bool {
+	for _, ref := range ownerRefs {
+		// UID can change if k8s rebuilds its object tree
+		// Kind and Name should be enough info to determine ownership conclusively
+		if ref.Kind == v1alpha1.ObjectBucketClaimGVK().Kind && ref.Name == obc.Name {
+			return true
+		}
+	}
+	return false
+}
+
+func bucketIsOwnedByClaim(obc *v1alpha1.ObjectBucketClaim, ob *v1alpha1.ObjectBucket) bool {
+	ref := ob.Spec.ClaimRef
+	emptyRef := corev1.ObjectReference{}
+	if ref == nil || *ref == emptyRef {
+		return false
+	}
+	// UID can change if k8s rebuilds its object tree
+	// Kind, Namespace, and Name should be enough info to determine ownership conclusively
+	if ref.Kind == v1alpha1.ObjectBucketClaimGVK().Kind && ref.Namespace == obc.Namespace && ref.Name == obc.Name {
+		return true
+	}
+	return false
+}
+
 func shouldProvision(obc *v1alpha1.ObjectBucketClaim) bool {
 	logD.Info("checking OBC for OB name, this indicates provisioning is complete", obc.Name)
-	if obc.Spec.ObjectBucketName != "" {
+	if obc.Spec.ObjectBucketName != "" && obc.Status.Phase == v1alpha1.ObjectBucketClaimStatusPhaseBound {
 		log.Info("provisioning already completed", "ObjectBucket", obc.Spec.ObjectBucketName)
 		return false
 	}
@@ -101,6 +127,14 @@ func isNewBucketByObjectBucket(c kubernetes.Interface, ob *v1alpha1.ObjectBucket
 		return false
 	}
 	return len(class.Parameters[v1alpha1.StorageClassBucket]) == 0
+}
+
+func composeConfigMapName(obc *v1alpha1.ObjectBucketClaim) string {
+	return obc.Name
+}
+
+func composeSecretName(obc *v1alpha1.ObjectBucketClaim) string {
+	return obc.Name
 }
 
 func configMapForClaimKey(key string, c kubernetes.Interface) (*corev1.ConfigMap, error) {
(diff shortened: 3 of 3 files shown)


BlaineEXE:
I think I just realized that this will help with the case where `bucketName` is given but will effectively remove a bucket that a user might have already been using if `generateBucketName` is set. I need to do more work on this.

BlaineEXE:
@thotz @leseb @travisn @guymguym @jeffvance @copejon please have a look now

BlaineEXE:
Update: I need to make another adjustment to handle the case for upgrading old claims with `generateBucketName` set. Older versions of the library set `bucketName` once a bucket was created for the claim, which is an error condition if provisioning runs again.

BlaineEXE:
I have tested upgrade scenarios with Rook from https://github.com/rook/rook/pull/6788 for both cases below with success:
- `bucketName` is set (`generateBucketName` is not set)
- `generateBucketName` is set (`bucketName` is unset)
  - there was a special upgrade case for the generate case because `bucketName` used to be set in addition once provisioning succeeded

I also did some manual testing where I injected `os.Exit` calls into the code and made sure that the changes to Secret and ConfigMap creation to make them idempotent worked as expected. During these tests I found that deleting the resource was not enough, and I had to delete the resource and wait for the resource to be deleted with a `Get` call.

pkg/provisioner/controller.go
```diff
@@ -291,6 +293,103 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+	obc, err = claimForKey(key, c.libClientset)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return fmt.Errorf("OBC was lost before we could provision: %v", err)
+		}
+		return err
+	}
+
+	// Older versions of this library set obc.Spec.BucketName with the name of the bucket generated
+	// from obc.Spec.GenerateBucketName, which causes an error condition when updating to a newer
```
thotz, line 306: AFAIR user do can specify both `bucketname` and `generatebucketname` in yaml file(may not be a valid case but still) and in that case `bucketname` has preference over `generatebucketname` while creating OBC. So this may result in a error

pkg/provisioner/resourcehandlers.go
```diff
@@ -53,6 +53,10 @@ const (
 	objectBucketNameFormat = "obc-%s-%s"
 )
 
+func configMapNameForClaim(obc *v1alpha1.ObjectBucketClaim) string {
+	return obc.Name
+}
+
```
thotz, line 59: Sorry I didn't understand use case of this function, we can directly access obc.Name in all its occurences. Please ignore it if is lack of my golang style of coding

pkg/provisioner/controller.go
```diff
@@ -470,7 +470,42 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		return fmt.Errorf("error getting reference to OBC: %v", err)
 	}
 
-	// create Secret and ConfigMap
+	// Create auth Secret and bucket info ConfigMap
+	// If the operator crashes before the OB is created, provisioning will happen again, and a
```
thotz, line 474: Just for the confirmation(to understand the code flow for me), OBC creation handled by the provisioner by Grant()/Provision(). Then request return back to library and then it creates cm or secret or both and before completing OB operator crashed, so we have dangling secrets or cm or both. When operator comes back it receives again on OBC claim, so operator need to hanlde Grant()/Provision() request gracefully without failing from its end.

Can we have similar situation Deletion() of OBC claim?

pkg/provisioner/resourcehandlers.go
BlaineEXE: I intend this just to make it a little clearer that the claim informs the name of the configmap and allows for it to change more easily in the future now that it is used in more than one place in the code.

pkg/provisioner/controller.go
BlaineEXE: I think your understanding sounds correct. I will update the comment to clarify "provisioning" on this line should be "Provision()/Grant()". 

I'm not quite sure I understand what you mean about deletion though...

pkg/provisioner/controller.go
```diff
@@ -291,6 +293,103 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	// to control access to static buckets via RBAC rules on storage classes.
 	isDynamicProvisioning := isNewBucketByStorageClass(class)
 
+	// Re-Get the claim in order to shorten the race condition where the claim was deleted after provisioning started
+	obc, err = claimForKey(key, c.libClientset)
+	if err != nil {
+		if errors.IsNotFound(err) {
```
jeffvance, line 299: Does this test imply that the finalzer just added above was not present? Otherwise the OBC could not actually be deleted. Are you seeing a window where we set the finalizer but it's not persisted and thus this GET is required?

pkg/provisioner/controller.go
```diff
@@ -291,6 +293,103 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+		return err
+	}
+
+	// Older versions of this library set obc.Spec.BucketName with the name of the bucket generated
+	// from obc.Spec.GenerateBucketName, which causes an error condition when updating to a newer
+	// version of this library if provisioning runs again. If the bucket name has the
+	// GenerateBucketName prefix, we can assume that this was originally intended to be a generated
+	// name.
+	if obc.Spec.GenerateBucketName != "" && obc.Spec.BucketName != "" {
+		if strings.HasPrefix(obc.Spec.BucketName, obc.Spec.GenerateBucketName+"-") {
```
jeffvance, line 311: I think the generation of a random bucket name should be made a helper func. I don't like that the algorithm for a creating a bucket name is found in 2 places in the code and thus can get out of sync.

pkg/provisioner/controller.go
```diff
@@ -291,6 +293,103 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+			ob.Spec.ClaimRef, err = claimRefForKey(key, c.libClientset)
+			if err != nil {
+				return fmt.Errorf("error getting reference to OBC: %v", err)
+			}
+			ob, err = updateObjectBucket(ob, c.libClientset, defaultRetryBaseInterval, defaultRetryTimeout)
+			if err != nil {
+				return fmt.Errorf("error updating object bucket with assumed claim ref: %v", err)
+			}
+		} else if obc.UID == ob.Spec.ClaimRef.UID {
+			logD.Info("found an existing Object Bucket belonging to this claim, re-associating this claim and stopping provisioning", "object bucket", ob.Name)
```
jeffvance, line 379: nit: why is "re-associating" in this err msg?

pkg/provisioner/controller.go
```diff
@@ -291,6 +293,103 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+		} else {
+			return fmt.Errorf("found an object bucket %q that is not owned by this object bucket claim, this is likely an artifact from a previous OBC that did not get cleaned up, not continuing with OBC provisioning", ob.Name)
```
jeffvance, line 381: The OBC's namespace is not shown in the err msg so we don't necessarily know which OBC is the problem.

pkg/provisioner/controller.go
```diff
@@ -291,6 +293,103 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+		}
+		err = updateOnSuccess()
+		if err != nil {
+			return fmt.Errorf("error re-associating object bucket to claim. %v", err)
+		}
+		// Do not do any more provisioning; the bucket is already provisioned and merely
+		// needed re-associated to the claim
+		logD.Info("re-associated bucket to claim successfully")
```
jeffvance, line 389: Most of the design doc for this lib refers to what you're calling re-association as _binding_. What do you think about s/re-association/binding/g ?

pkg/provisioner/controller.go
```diff
@@ -291,6 +293,103 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+			return fmt.Errorf("error updating OBC: %v", err)
+		}
+		obc, err = updateObjectBucketClaimPhase(
+			c.libClientset,
+			obc,
+			v1alpha1.ObjectBucketClaimStatusPhaseBound,
+			defaultRetryBaseInterval,
+			defaultRetryTimeout)
+		if err != nil {
+			return fmt.Errorf("error updating OBC %q's status to: %v", v1alpha1.ObjectBucketClaimStatusPhaseBound, err)
```
jeffvance, line 356: does it matter here and the `return` above that the OB has it's Phase set to "Bound"?

pkg/provisioner/controller.go
```diff
@@ -291,6 +293,103 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+		}
+		return nil
+	}
+
+	// recover from degraded state where provisioning is happening again on a completed obc
+	ob, err = getObjectBucketFromClaimKey(key, c.libClientset, defaultRetryBaseInterval, defaultRetryTimeout)
+	if err != nil {
+		return fmt.Errorf("error determining if object bucket for this claim exists: %v", err)
+	}
+	if ob != nil {
```
jeffvance, line 366: A bit simplified:
```
	if ob != nil {
...		if ob.Spec.ClaimRef == nil || *ob.Spec.ClaimRef == emptyRef || ob.Spec.ClaimRef.UID == types.UID("") {
			logD.Info("Attempting to re-bind OB %q with OBC %s/%s", ob.Name, obc.Namespace, obc.Name)
			ob.Spec.ClaimRef, err = claimRefForKey(key, c.libClientset)
			if err != nil {
				return fmt.Errorf("error getting OB's reference to OBC: %v", err)
			}
			ob, err = updateObjectBucket(ob, c.libClientset, defaultRetryBaseInterval, defaultRetryTimeout)
			if err != nil {
				return fmt.Errorf("error setting OB's reference to OBC: %v", err)
			}
		} else if obc.UID != ob.Spec.ClaimRef.UID {
			return fmt.Errorf("OB %q that is not owned by this OBC, which is likely an artifact from a previous OBC that wasn't cleaned up. Stopping OBC provisioning", ob.Name)
		}
		err = updateOnSuccess()
		if err != nil {
			return fmt.Errorf("error re-binding OB %q with OBC %s/%s", ob.Name, obc.Namespace, obc.Name, err)
		}
		// Do not do any more provisioning; the bucket is already provisioned and merely
		// needed re-associated to the claim
		logD.Info("re-binding OB to OBC successful")
		return nil
```

pkg/provisioner/controller.go
```diff
@@ -291,6 +293,103 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+		}
+		return err
+	}
+
+	// Older versions of this library set obc.Spec.BucketName with the name of the bucket generated
+	// from obc.Spec.GenerateBucketName, which causes an error condition when updating to a newer
+	// version of this library if provisioning runs again. If the bucket name has the
+	// GenerateBucketName prefix, we can assume that this was originally intended to be a generated
+	// name.
+	if obc.Spec.GenerateBucketName != "" && obc.Spec.BucketName != "" {
```
jeffvance, line 310: The design doc allows both of these names to be specified and states in that case `bucketName` has precedence over `generateBucketName`. So, unless you have API validation or an admission controller to remove `bucketName`, when `generateBucketName` is specified, this 'if` will blank out bucket name and thus change the hierarchy described in the doc. This is a backwards incompatible change, if I understand it correctly.

pkg/provisioner/controller.go
```diff
@@ -380,7 +470,43 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		return fmt.Errorf("error getting reference to OBC: %v", err)
```
jeffvance, line 470: not your problem, but the obc's namespace is missing from the err msg

pkg/provisioner/controller.go
```diff
@@ -380,7 +470,43 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
 	}
 
-	// create Secret and ConfigMap
+	// Create auth Secret and bucket info ConfigMap
+	// If the operator crashes before the OB is created, provisioning will happen again including a
+	// call to Provision()/Grant(), and a Secret and/or ConfigMap might exist from the previous
+	// provisioning attempt. Provision() isn't guaranteed to give the same credentials or bucket
+	// each time, so if the Secret/ConfigMap already exists, delete the old one and re-create it
+	// with the latest info.
+
```
jeffvance, line 479: Is it a safe assumption that n the above scenario (retries on partially provisioned obc) that the provisioner may have generated artifacts, eg. users? How does a provisioner know to cleanup these artifacts if it doesn't get a _Delete_ or _Revoke_ call, or if the OB wasn't updated with the provisioner's shared state info?

pkg/provisioner/resourcehandlers.go
```diff
@@ -176,6 +225,106 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, lab
 	return configMap, err
 }
 
+func getExistingSecret(obc *v1alpha1.ObjectBucketClaim, c kubernetes.Interface, retryInterval, retryTimeout time.Duration) (secret *corev1.Secret, err error) {
+	err = wait.PollImmediate(retryInterval, retryTimeout, func() (done bool, err error) {
+		secret, err = c.CoreV1().Secrets(obc.Namespace).Get(context.TODO(), secretNameForClaim(obc), metav1.GetOptions{})
+		if err != nil {
+			if errors.IsNotFound(err) {
+				// Not found is valid here which returns no error and a nil secret
+				secret = nil
```
jeffvance, line 234: secret and err are already nil since that is their "zero" (default) value.

pkg/provisioner/resourcehandlers.go
```diff
@@ -176,6 +225,106 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, lab
 	return configMap, err
 }
 
+func getExistingSecret(obc *v1alpha1.ObjectBucketClaim, c kubernetes.Interface, retryInterval, retryTimeout time.Duration) (secret *corev1.Secret, err error) {
+	err = wait.PollImmediate(retryInterval, retryTimeout, func() (done bool, err error) {
```
jeffvance, line 229: I'm not sure we need polling here. The controller will retry when an error is returned to some max number of attempts. So, I think it's ok to not poll, return an error, even if it's an intermittent error and retry naturally? But, I have not gone through all the code to check idempotency.

pkg/provisioner/controller.go
```diff
@@ -380,7 +470,43 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+	secret, err = getExistingSecret(obc, c.clientset, defaultRetryBaseInterval, defaultRetryTimeout)
+	if err != nil {
+		return fmt.Errorf("error checking for existing Secret for OBC: %v", err)
```
jeffvance, line 482: should the obc namespace/name be inserted into the msg?

pkg/provisioner/controller.go
```diff
@@ -380,7 +470,43 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+	}
+	if secret != nil {
+		if !isOwnedByClaim(obc, secret.OwnerReferences) {
+			return fmt.Errorf("found an existing OBC credentials Secret that is not owned by this OBC, this may be an artifact from a previous OBC that did not get cleaned up or a user-created resource, not continuing with OBC provisioning")
```
jeffvance, line 486: Isn't it possible that the secret was created but the ownership part, which is set in a separate step, is where we crashed? In this case the secret is correct but only partially filled in. Since secrets are immutable we cannot edit it so I agree with erroring out. I'd like to see the error msg suggest to delete the secret.

pkg/provisioner/controller.go
```diff
@@ -380,7 +470,43 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
 
-	// create Secret and ConfigMap
+	// Create auth Secret and bucket info ConfigMap
+	// If the operator crashes before the OB is created, provisioning will happen again including a
+	// call to Provision()/Grant(), and a Secret and/or ConfigMap might exist from the previous
+	// provisioning attempt. Provision() isn't guaranteed to give the same credentials or bucket
+	// each time, so if the Secret/ConfigMap already exists, delete the old one and re-create it
+	// with the latest info.
+
+	secret, err = getExistingSecret(obc, c.clientset, defaultRetryBaseInterval, defaultRetryTimeout)
```
jeffvance, line 480: Can all of this secret and CM handling be contained in helper funcs, just to improve readability by shortening this func?

pkg/provisioner/resourcehandlers.go
```diff
@@ -53,6 +53,10 @@ const (
 	objectBucketNameFormat = "obc-%s-%s"
 )
 
+func configMapNameForClaim(obc *v1alpha1.ObjectBucketClaim) string {
```
jeffvance, line 56: 1) why is this not in helpers? 2) why isn't there a parallel func for secret names? 3) the func name pattern is used through the lib but in all case but here the func returns is a pointer to the resource and an error. Therfore, I think a different func name is beneficial, eg. `generateConfigMapName` or `makeCMName` etc.

pkg/provisioner/resourcehandlers.go
```diff
@@ -84,6 +88,10 @@ func newBucketConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint,
 	}, nil
 }
 
+func secretNameForClaim(obc *v1alpha1.ObjectBucketClaim) string {
```
jeffvance, line 91: same comments as for CM.

pkg/provisioner/resourcehandlers.go
```diff
@@ -131,6 +139,47 @@ func createObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface, retryI
 	return
 }
 
+func updateObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface, retryInterval, retryTimeout time.Duration) (result *v1alpha1.ObjectBucket, err error) {
```
jeffvance, line 142: missing func comment

pkg/provisioner/resourcehandlers.go
```diff
@@ -131,6 +139,47 @@ func createObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface, retryI
...
+	logD.Info("updating ObjectBucket", "name", ob.Name)
+
+	err = wait.PollImmediate(retryInterval, retryTimeout, func() (bool, error) {
```
jeffvance, line 145: my same polling question/comment applies here.

pkg/provisioner/resourcehandlers.go
```diff
@@ -131,6 +139,47 @@ func createObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface, retryI
...
+		result, err = c.ObjectbucketV1alpha1().ObjectBuckets().Update(context.TODO(), ob, metav1.UpdateOptions{})
+		if err != nil {
+			// could be intermittent api error
+			log.Error(err, "probably not fatal, retrying")
+			return false, nil
+		}
+		return true, nil // update successful
+	})
+	return result, err
```
jeffvance, line 154: in Go since you've named the function returns, you can simply end the func with `return` (or maybe return isn't even required?)

pkg/provisioner/resourcehandlers.go
```diff
@@ -131,6 +139,47 @@ func createObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface, retryI
...
+}
+
+func getObjectBucketFromClaimKey(key string, c versioned.Interface, retryInterval, retryTimeout time.Duration) (result *v1alpha1.ObjectBucket, err error) {
+	obName, err := objectBucketNameFromClaimKey(key)
+	if err != nil {
+		return nil, err
+	}
+
+	logD.Info("seeing if object bucket for claim exists", "expected object bucket name", obName)
+	err = wait.PollImmediate(retryInterval, retryTimeout, func() (bool, error) {
```
jeffvance, line 164: ditto on polling

pkg/provisioner/resourcehandlers.go
```diff
@@ -131,6 +139,47 @@ func createObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface, retryI
...
+			log.Error(err, "probably not fatal, retrying")
+			return false, nil
+		}
+		return true, nil // update successful
+	})
+	return result, err
+}
+
+func getObjectBucketFromClaimKey(key string, c versioned.Interface, retryInterval, retryTimeout time.Duration) (result *v1alpha1.ObjectBucket, err error) {
+	obName, err := objectBucketNameFromClaimKey(key)
```
jeffvance, line 158: Note: this `err` creates a new (shadow) err separate from your return value

pkg/provisioner/resourcehandlers.go
```diff
@@ -176,6 +225,106 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, lab
...
+func deleteSecret(obc *v1alpha1.ObjectBucketClaim, c kubernetes.Interface, retryInterval, retryTimeout time.Duration) error {
+	deleteFunc := func() error {
+		return c.CoreV1().Secrets(obc.Namespace).Delete(context.TODO(), secretNameForClaim(obc), metav1.DeleteOptions{})
+	}
+	getFunc := func() error {
+		_, err := c.CoreV1().Secrets(obc.Namespace).Get(context.TODO(), secretNameForClaim(obc), metav1.GetOptions{})
+		return err
+	}
+	logD.Info("deleting secret", "secret name", secretNameForClaim(obc))
+	return deleteResourceAndWait(deleteFunc, getFunc, retryInterval, retryTimeout)
```
jeffvance, line 277: Nice!

pkg/provisioner/resourcehandlers.go
```diff
@@ -176,6 +225,106 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, lab
...
+		_, err := c.CoreV1().ConfigMaps(obc.Namespace).Get(context.TODO(), configMapNameForClaim(obc), metav1.GetOptions{})
+		return err
+	}
+	logD.Info("deleting configmap", "cm name", configMapNameForClaim(obc))
+	return deleteResourceAndWait(deleteFunc, getFunc, retryInterval, retryTimeout)
+}
+
+func deleteResourceAndWait(deleteFunc func() error, getFunc func() error, retryInterval, retryTimeout time.Duration) error {
+	var err error
+	err = wait.PollImmediate(retryInterval, retryTimeout, func() (done bool, err error) {
```
jeffvance, line 294: ditto on polling

pkg/provisioner/resourcehandlers.go
```diff
@@ -176,6 +225,106 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, lab
...
+	getFunc := func() error {
+		_, err := c.CoreV1().ConfigMaps(obc.Namespace).Get(context.TODO(), configMapNameForClaim(obc), metav1.GetOptions{})
+		return err
+	}
+	logD.Info("deleting configmap", "cm name", configMapNameForClaim(obc))
+	return deleteResourceAndWait(deleteFunc, getFunc, retryInterval, retryTimeout)
+}
+
+func deleteResourceAndWait(deleteFunc func() error, getFunc func() error, retryInterval, retryTimeout time.Duration) error {
+	var err error
```
jeffvance, line 293: could just name the returned error

pkg/provisioner/resourcehandlers.go
```diff
@@ -176,6 +225,106 @@ func createConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, lab
...
+			log.Error(err, "probably not fatal, retrying")
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		return err
+	}
+	err = wait.PollImmediate(retryInterval, retryTimeout, func() (done bool, err error) {
+		err = getFunc()
```
jeffvance, line 312: Is the point of the get func to wait in case the resource has a finalizer? Otherwise, since you are not returning the resource in cases of Delete failure, I don't see the value in get. Also, a top func comment would help here.

pkg/provisioner/controller.go
BlaineEXE: This code block I did not add but merely moved up from below. Maybe this block isn't necessary and I can remove it?

pkg/provisioner/controller.go
BlaineEXE: This is difficult because later in the code, it is an error condition if both are set, and no provisioning will occur. I could alternately fix this by changing the behavior to be as you say and allow both to be set but to pick `bucketName` as preferred.

pkg/provisioner/controller.go
BlaineEXE: It seems like the code as implemented and the design doc are out of sync. See the code snippet below which I have not modified.

```go
func composeBucketName(obc *v1alpha1.ObjectBucketClaim) (string, error) {
	if obc.Spec.BucketName == "" && obc.Spec.GenerateBucketName == "" {
		return "", fmt.Errorf("expected either bucketName or generateBucketName defined")
	}
	if obc.Spec.BucketName != "" && obc.Spec.GenerateBucketName != "" {
		return "", fmt.Errorf("cannot define both bucketName and generateBucketName")
	}
	bucketName := obc.Spec.BucketName
	if bucketName == "" {
		bucketName = generateBucketName(obc.Spec.GenerateBucketName)
	}
	return bucketName, nil
}
```

GitLens in my editor shows that the line of behavior I have been taking to be canon was added by you.
```go
	if obc.Spec.BucketName != "" && obc.Spec.GenerateBucketName != "" {
		return "", fmt.Errorf("cannot define both bucketName and generateBucketName")
	}
```
```
allow obc to have no bucket related fields for "brownfield" (#88)

* change ob ClaimRef from UID to ObjectReference

* added UID to ob.ClaimRef

* allow obc to have no bkt name fields for brownfield

```

Should I instead drop this recovery code and change the behavior to prefer `bucketName` over `generateBucketName`?

pkg/provisioner/controller.go
BlaineEXE: I'm not quite following the question. I'm sorry.

pkg/provisioner/controller.go
BlaineEXE: Good point. I'll fix that.

pkg/provisioner/controller.go
BlaineEXE: The re-associating happens below with `updateOnSuccess()`. The message is for users (or more realistically us when we are debugging issues).

pkg/provisioner/controller.go
BlaineEXE: That makes sense. I'll change to `re-bind` instead.

pkg/provisioner/controller.go
BlaineEXE: It is, and multiple user creation happens with provision crash loops in the code before #198. 

TBH, because the Provisioner implements this, I don't see a good way to fix this issue right now. In Ceph we can't really know if a user was created specifically for an OBC or was created by a user, so it's not a trivial task if the provisioner says "clean up now." We would need to know what users are in use by *all* OBCs and have a guarantee from the controller and all its workers that it will halt all operations while Ceph searches its users to remove unknown ones while also doing its best to not remove users for its own CRDs or that might have been created by the user on the CLI.

pkg/provisioner/controller.go
BlaineEXE: If we error out, we cause infinite provisioning loops, which is what was the root cause of the infinite provision loop fixed in #198 and can cause cleanup and data loss now that #198 is in. The secret needs to be able to be updated so this cannot happen.

pkg/provisioner/resourcehandlers.go
BlaineEXE: 1. I'll move that no problem
2. I'm not sure what you are suggesting here. `composeBucketName` returns a string and error, but this function doesn't need to have an error return.

pkg/provisioner/resourcehandlers.go
BlaineEXE: Functions that are not exported in `Go` don't require comments, and it hasn't been the trend in this library from what I've seen to add comments to all functions regardless. I can add one here, but it is going against trend, and I don't think one is needed.

pkg/provisioner/resourcehandlers.go
BlaineEXE: I am aware and also prefer the explicit return for readability.

pkg/provisioner/resourcehandlers.go
BlaineEXE: I do agree I think. It's a matter of whether polling/recovery is better in the face of random API errors or erroring out to go through the provisioning loop again. I suppose it's most important to handle the random errors when creating/updating resources IMO. A pro of having a polling loop in functions even like this does mean that the provision function can be a more stable atomic unit, and I do see polling loops used a lot throughout the code as it stands already.

pkg/provisioner/controller.go
jeffvance: I'm guessing the original location for this block made more sense. What is your reason for moving it? I am not sure it's even needed now but haven't checked more deeply to figure that out yet...

pkg/provisioner/controller.go
jeffvance: That's a good question. The code is not necessarily wrong but for sure the code and doc are not in sync here. I lean toward creating a small issue against the doc and leaving the code as is.

pkg/provisioner/controller.go
jeffvance: What I noticed is that we 1) update ob's Phase to "Bound", 2) update the obc with the binding OB.name, 3) update the obc's Phase to "Bound".  Does it matter if 1) and 2) or just 1) succeeds but the subsequent updates fail? IOW, assume 1 succeeds and 2 fails: we now have the OB saying bound but the OBC appears not bound. So we return an error and retry which I guess is fine. This comment was more about thinking out loud apparently...

pkg/provisioner/controller.go
jeffvance: The main point of this comment above, besides a bit shorter messages, if that I don't see the need for the `} else if obc.UID == ob.Spec.ClaimRef.UID {`

pkg/provisioner/controller.go
jeffvance: Ack. This is a tough problem. COSI decided that all provisioners _must_ be idempotent but has no provisioner "cleanup" mechanism, so I think we'll see the same issue in COSI.

pkg/provisioner/controller.go
jeffvance: Maybe I'm missing something here. 1) When I wrote "error out" I was referring to L486 and L491 so I don't understand your reply of "_If we error out, we cause infinite provisioning loops_" because we are error-ing out (at least by me definition :-). 2) You wrote "_The secret needs to be able to be updated_", but Secrets are immutable and the code handles this by deleting the Secret and creating a new one. 
The interesting case here is when the secret name matches what the lib wants to create but the ownership doesn't match, and in that case I was suggesting the error msg suggest that this unexpected secret be deleted.

pkg/provisioner/resourcehandlers.go
jeffvance: Yes regarding `composeBucketName` but my point above is that composeBucketName is not named `bucketNameFromClaim` -- this is the pattern I'm referring to. It's a nit but I noticed it.

pkg/provisioner/resourcehandlers.go
jeffvance: Agree we used polling throughout the lib code (but, unfortunately, not consistently. We may have not understood controllers well enough back then? I do agree that these polls make the func more "atomic" and likely hide all intermittend k8s api errors. Still, it's extra code and slightly impairs readability... Probably fine in this PR to continue the lib's use of Poll and, if it ever matters, handle it in a separate PR.

pkg/provisioner/controller.go
BlaineEXE: Should I move this to before `setOBCMetaFields`? Remove it? I'll follow your recommendation on this.

pkg/provisioner/controller.go
BlaineEXE: If the code gives precedence to `bucketName` as in the doc, that makes the upgrade recovery case easier. We would no longer have to infer that a `bucketName` that has `generateBucketName` as a prefix should originally have been only `generateBucketName`. I think maybe this is a slightly better option.

pkg/provisioner/controller.go
BlaineEXE: I don't think there are any functional issues other than OB displaying "Bound" before OBC reports that. FWIW, existing code in Rook has an infinite OBC provision loop which leaves OB to "Bound" but resets "OBC" to "Pending." Maybe the code that resets "OBC" to "Pending" should also set "OB" to "Pending"?

pkg/provisioner/resourcehandlers.go
BlaineEXE: The `Get` waits for the resource to be deleted. A new resource can't be created unless the old one is already gone.

pkg/provisioner/controller.go
BlaineEXE: Okay. I thought there was more intent other than suggesting users delete the secret (or configmap in that case).

pkg/provisioner/controller.go
BlaineEXE: I'd like to keep the middle `else if` line in order to have a definite log message show when the library is taking the code path where it is updating an OBC to re-bind to an existing OB for the upgrade case.

pkg/provisioner/resourcehandlers.go
BlaineEXE: I took your suggestion to remove polling, but I think it still makes sense here as this comprises the "wait" portion of "deleteAndWait".

pkg/provisioner/controller.go
BlaineEXE: This is a pretty unlikely corner case. I would anticipate it happening only on node failure/partition or similar events while provisioning is ongoing.

pkg/provisioner/controller.go
BlaineEXE: I didn't take your suggestion word-for-word where I thought the info might be able to be refined a little. I did update the messages and made sure the OB and OBC info would get reported. LMK if you still have changes.

pkg/provisioner/controller.go
jeffvance: Comment resolved.

> Maybe the code that resets "OBC" to "Pending" should also set "OB" to "Pending"?

This seems worthwhile to me.

BlaineEXE:
For reference, I tested idempotency of this PR with the following commit: https://github.com/BlaineEXE/lib-bucket-provisioner/commit/e9462e2ad742d8a909cf1aabbfa308cd6e5b642d

pkg/provisioner/resourcehandlers.go
jeffvance: True!

pkg/provisioner/controller.go
```diff
@@ -302,13 +302,72 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
 	}
 
-	// Re-Get the claim in order to shorten the race condition where the claim was deleted after provisioning started
-	obc, err = claimForKey(key, c.libClientset)
+	// In the case where a bucket name is being generated, generate the name and store it in the OBC
+	// spec before doing any Provisioning so that any crashes encountered in this code will not
+	// result in multiple buckets being generated for the same OBC. bucketName takes precedence over
+	// generateBucketName if both are present.
+	if obc.Spec.BucketName == "" {
+		obc.Spec.BucketName = bucketName
```
jeffvance, line 310: For API efficiency, is it acceptable to update the OBC inside of `updateOnSuccess`?

pkg/provisioner/controller.go
jeffvance: Looking at the latest code, you've decided to forgo the log msgs related to re-binding and claimRef/UID checks. Is this correct of has this previous block been moved?

pkg/provisioner/controller.go
```diff
@@ -380,7 +439,18 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		return fmt.Errorf("error getting reference to OBC: %v", err)
 	}
 
-	// create Secret and ConfigMap
+	// Create auth Secret and bucket info ConfigMap
```
jeffvance, line 442: `setBasicInfo` is called [here](https://github.com/kube-object-storage/lib-bucket-provisioner/pull/200/files#diff-1e346e3430c7a1bec9a1fde398367f33b921671c6c3ced081ad42e4d8e345522R400). What is the reason for also calling it in the [`defer`](https://github.com/kube-object-storage/lib-bucket-provisioner/pull/200/files#diff-1e346e3430c7a1bec9a1fde398367f33b921671c6c3ced081ad42e4d8e345522R400)? Is this to handle cases of the provisioner unsetting some of the ob fields?  

BTW, I know this pr does not change this code, but I noticed it in my review.

pkg/provisioner/resourcehandlers.go
```diff
@@ -131,6 +131,53 @@ func createObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface, retryI
...
+	return result, err
+}
+
+func getObjectBucketFromClaimKey(key string, obc *v1alpha1.ObjectBucketClaim, c versioned.Interface, retryInterval, retryTimeout time.Duration) (ob *v1alpha1.ObjectBucket, err error) {
+	obName, err := objectBucketNameFromClaimKey(key)
+	if err != nil {
+		return nil, err
+	}
+
+	logD.Info("seeing if object bucket for claim exists", "expected object bucket name", obName)
```
jeffvance, line 149: maybe clearer?
`	logD.Info("seeing if OB for OBC exists", "checking for OB name", obName)`

pkg/provisioner/resourcehandlers.go
```diff
@@ -131,6 +131,53 @@ func createObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface, retryI
...
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	emptyRef := corev1.ObjectReference{}
+	if ob.Spec.ClaimRef == nil || *ob.Spec.ClaimRef == emptyRef {
+		// If the CRD validation for the OB improperly specifies the object reference field, it will
+		// be missing. Do our best to recover from this case gracefully.
+		logD.Info("found OB matching this OBC that does not have an object reference, assuming it should belong to this claim", "OB", ob.Name)
```
jeffvance, line 163: "object reference" seems vague to me. How about:
`logD.Info("found OB matching this OBC name but the OB's ClaimRef is empty. Binding OB", ob.Name, "to OBC", obc.Name)`

pkg/provisioner/resourcehandlers.go
```diff
@@ -131,6 +131,53 @@ func createObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface, retryI
...
+			return nil, fmt.Errorf("error updating OB %q with assumed claim ref: %v", ob.Name, err)
+		}
+	} else if bucketIsOwnedByClaim(obc, ob) {
+		logD.Info("found OB belonging to this OBC", "OB", ob.Name)
+	} else {
+		return nil, fmt.Errorf("found OB %q not owned by OBC %q, likely an artifact from a previous OBC that did not get cleaned up, user must delete the OB in order to allow OBC provisioning to continue", ob.Name, key)
+	}
+
+	return ob, err
+}
```
jeffvance, line 179: This is a nit but I'd restructure as:
```
...
   if bucketIsOwnedByClaim(obc, ob) {
         logD.Info(sprintf("found OB %q matching OBC %s/%s but the OB's ClaimRef is empty. Binding OB to OBC", ob.Name, obc.Namespace, obc.Name))
         return ob, err
   }

   if ob.Spec.ClaimRef == nil || *ob.Spec.ClaimRef == emptyRef {
		// If the CRD validation for the OB improperly specifies the object reference field, it will
		// be missing. Do our best to recover from this case gracefully.
		logD.Info("found OB matching this OBC that does not have an object reference, assuming it should belong to this claim", "OB", ob.Name)
		ob.Spec.ClaimRef, err = claimRefForKey(key, c)
		if err != nil {
		      err = fmt.Sprintf("error generating OB %q's reference to OBC: %v", ob.Name, err)
		} else
		       ob, err = updateObjectBucket(ob, c, defaultRetryBaseInterval, defaultRetryTimeout)
		             if err != nil {
			            err = fmt.Sprintf("error updating OB %q with assumed claim ref: %v", ob.Name, err)
		             }
   } else {
          err = fmt.Sprintf("found OB %q not owned by OBC %q, likely an artifact from a previous OBC that did not get cleaned up, user must delete the OB in order to allow OBC provisioning to continue: %v", ob.Name, key, err) 
   }
   return nil, err
}
```

pkg/provisioner/resourcehandlers.go
jeffvance: agree

jeffvance:
Looks good @BlaineEXE . I just have a couple of mostly nit questions.

pkg/provisioner/controller.go
BlaineEXE: The ob from the provisioner can come back without info about the storage class, but the cleanup 'Delete' will need that info. Basically, just fill out the known info more to eliminate problems cleaning up caused by lack of that info.

pkg/provisioner/resourcehandlers.go
BlaineEXE: Fixing without %s/%s.`logD` automatically prints the key which identifies the OBC namespace/name.

pkg/provisioner/resourcehandlers.go
BlaineEXE: The restructure you suggested mixed up the logic a bit, but I restructured in the spirit of your comment.

jeffvance: In thinking about this more last night, I'm not sure I like it anymore. I was hoping to reduce the number or returns but I don't want to mess up the logic. Feel free to ignore!

jeffvance:
LGTM

PR #201, opened 2020-12-18
Handle ob/obc=nil when clean up provision failures

BlaineEXE:
If there are failures during provisioning, the OB or OBC can be nil
    when the cleanup function runs on an error. Allow both OB and OBC
    to keep their previous info if create/update of OB/OBC fails. If the
    create/update succeeds, fill in the OB/OBC info with the latest
    successful info.

    If OB/OBC become nil on create/update before cleanup, it can result in
    the provisioner's Delte() call failing to completely clean up resources
    from the failed Provision().

    The setBasicOBInfo function called after provisioning and before the
    deferred cleanup ops was also making debugging this issue more mentally
    cumbersome. Revise code flow to still update the necessary OB
    information needed for cleanup before returning errors from
    provisioning. This does not change behavior but makes the code flow more
    conceptually linear.

### DIFF
--- a/pkg/provisioner/controller.go
+++ b/pkg/provisioner/controller.go
@@ -252,9 +252,7 @@ func (c *obcController) syncHandler(key string) error {
 	obc, err = updateObjectBucketClaimPhase(
 		c.libClientset,
 		obc,
-		v1alpha1.ObjectBucketClaimStatusPhasePending,
-		defaultRetryBaseInterval,
-		defaultRetryTimeout)
+		v1alpha1.ObjectBucketClaimStatusPhasePending)
 	if err != nil {
 		return fmt.Errorf("error updating OBC status: %s", err)
 	}
@@ -310,9 +308,7 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		obc.Spec.BucketName = bucketName
 		obc, err = updateClaim(
 			c.libClientset,
-			obc,
-			defaultRetryBaseInterval,
-			defaultRetryTimeout)
+			obc)
 		if err != nil {
 			return fmt.Errorf("error updating OBC %q with bucket name: %v", key, err)
 		}
@@ -325,9 +321,7 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		ob, err = updateObjectBucketPhase(
 			c.libClientset,
 			ob,
-			v1alpha1.ObjectBucketStatusPhaseBound,
-			defaultRetryBaseInterval,
-			defaultRetryTimeout)
+			v1alpha1.ObjectBucketStatusPhaseBound)
 		if err != nil {
 			return fmt.Errorf("error updating OB %q's status to %q: %v", ob.Name, v1alpha1.ObjectBucketStatusPhaseBound, err)
 		}
@@ -336,18 +330,14 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		obc.Spec.BucketName = bucketName
 		obc, err = updateClaim(
 			c.libClientset,
-			obc,
-			defaultRetryBaseInterval,
-			defaultRetryTimeout)
+			obc)
 		if err != nil {
 			return fmt.Errorf("error updating OBC: %v", err)
 		}
 		obc, err = updateObjectBucketClaimPhase(
 			c.libClientset,
 			obc,
-			v1alpha1.ObjectBucketClaimStatusPhaseBound,
-			defaultRetryBaseInterval,
-			defaultRetryTimeout)
+			v1alpha1.ObjectBucketClaimStatusPhaseBound)
 		if err != nil {
 			return fmt.Errorf("error updating OBC %q's status to: %v", v1alpha1.ObjectBucketClaimStatusPhaseBound, err)
 		}
@@ -377,27 +367,11 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		Parameters:        class.Parameters,
 	}
 
-	// Fill in basic necessary object bucket info
-	setBasicOBInfo := func() {
-		setObjectBucketName(ob, key)
-		ob.Spec.StorageClassName = obc.Spec.StorageClassName
-		if ob.Spec.ReclaimPolicy == nil || *ob.Spec.ReclaimPolicy == corev1.PersistentVolumeReclaimPolicy("") {
-			// Do not blindly overwrite the reclaim policy. The provisioner might have reason to
-			// specify a reclaim policy that is  different from the storage class.
-			ob.Spec.ReclaimPolicy = options.ReclaimPolicy
-		}
-		ob.SetLabels(c.provisionerLabels)
-	}
-
 	// Should an error be returned, attempt to clean up the object store and API servers by
 	// calling the appropriate provisioner method.  In cases where Provision() or Revoke()
 	// return an err, it's likely that the ob == nil, hindering cleanup.
 	defer func() {
 		if err != nil && ob != nil {
-			// If provisioning fails before basic OB info is applied, cleanup below can fail because
-			// provisioner methods might not have access to information (like StorageClassName)
-			// they need to properly clean up. Therefore, we must set basic OB info here.
-			setBasicOBInfo()
 			log.Info("cleaning up provisioning artifacts")
 			if /*greenfield*/ isDynamicProvisioning && !pErr.IsBucketExists(err) {
 				log.Info("deleting provisioned resources")
@@ -425,14 +399,33 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	} else {
 		ob, err = c.provisioner.Grant(options)
 	}
+	// Record whether the provisioner returned an empty object bucket for error handling use later
+	emptyBucket := (ob == nil || ob == (&v1alpha1.ObjectBucket{}))
+
+	// Set information about the OB as soon as possible, even before handling errors so that the
+	// deferred cleanup function is not missing critical information needed to clean up.
+	// More note on above: can't check if provisioner returned empty OB after we set OB info below.
+	// The returned OB should only be nil if there is also an error with provisioning; in such a
+	// case, the provisioner should have cleaned up after itself already so we do not need to set
+	// info for OB cleanup.
+	if ob != nil {
+		setObjectBucketName(ob, key)
+		ob.Spec.StorageClassName = obc.Spec.StorageClassName
+		if ob.Spec.ReclaimPolicy == nil || *ob.Spec.ReclaimPolicy == corev1.PersistentVolumeReclaimPolicy("") {
+			// Do not blindly overwrite the reclaim policy. The provisioner might have reason to
+			// specify a reclaim policy that is  different from the storage class.
+			ob.Spec.ReclaimPolicy = options.ReclaimPolicy
+		}
+		ob.SetLabels(c.provisionerLabels)
+	}
+
 	if err != nil {
 		return fmt.Errorf("error %s bucket: %v", verb, err)
-	} else if ob == (&v1alpha1.ObjectBucket{}) {
+	} else if emptyBucket {
 		return fmt.Errorf("provisioner returned empty object bucket")
 	}
 
-	// Fill in known information in the object bucket struct as soon as possible
-	setBasicOBInfo()
+	// This info set on the OB isn't critical for cleanup
 	ob.SetFinalizers([]string{finalizer})
 	ob.Spec.ClaimRef, err = claimRefForKey(key, c.libClientset)
 	if err != nil {
@@ -521,7 +514,7 @@ func (c *obcController) handleDeleteClaim(key string, obc *v1alpha1.ObjectBucket
 
 	// call Delete or Revoke and then delete generated k8s resources
 	// Note: if Delete or Revoke return err then we do not try to delete resources
-	ob, err := updateObjectBucketPhase(c.libClientset, ob, v1alpha1.ObjectBucketClaimStatusPhaseReleased, defaultRetryBaseInterval, defaultRetryTimeout)
+	ob, err := updateObjectBucketPhase(c.libClientset, ob, v1alpha1.ObjectBucketClaimStatusPhaseReleased)
 	if err != nil {
 		return err
 	}
@@ -623,7 +616,7 @@ func (c *obcController) setOBCMetaFields(obc *v1alpha1.ObjectBucketClaim) (err e
 	obc.SetLabels(c.provisionerLabels)
 
 	logD.Info("updating OBC metadata")
-	obc, err = updateClaim(clib, obc, defaultRetryBaseInterval, defaultRetryTimeout)
+	obc, err = updateClaim(clib, obc)
 	if err != nil {
 		return fmt.Errorf("error configuring obc metadata: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


jeffvance:
Why not also check for nil in addition to empty [here](https://github.com/kube-object-storage/lib-bucket-provisioner/blob/5875d492f11115fb3a270bb42f3d49bdde8e4d83/pkg/provisioner/controller.go#L444)? Would this eliminate the need for `createdOB` in `updateOnSuccess`?

pkg/provisioner/controller.go
```diff
@@ -331,18 +334,24 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		if err != nil {
 			return fmt.Errorf("error updating OB %q's status to %q: %v", ob.Name, v1alpha1.ObjectBucketStatusPhaseBound, err)
 		}
+		ob = createdOB // don't let ob become nil but allow it to be updated on success
+
 		// update OBC
+		// Note: just as with the ob, don't all obc to be nil if returning for error for cleanup
```
jeffvance, line 340: do you mean "..., don't allow..."?

pkg/provisioner/controller.go
```diff
@@ -331,18 +334,24 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+		// because obc has important information also
```
jeffvance, line 341: how does the obc become nil?

pkg/provisioner/controller.go
```diff
@@ -351,6 +360,8 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 		if err != nil {
 			return fmt.Errorf("error updating OBC %q's status to: %v", v1alpha1.ObjectBucketClaimStatusPhaseBound, err)
 		}
+		obc = updatedOBC // don't let obc become nil but allow it to be updated on success
```
jeffvance, line 363: this assignment has already been done above and seems unnecessary here.

pkg/provisioner/controller.go
```diff
@@ -380,7 +391,10 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	// Fill in basic necessary object bucket info
 	setBasicOBInfo := func() {
 		setObjectBucketName(ob, key)
-		ob.Spec.StorageClassName = obc.Spec.StorageClassName
+		// obc can be nil if functions that return a modified version of it fail
```
jeffvance, line 394: why? If a func the returns an obc fails why does the code continue with that obc? err != nil should returns an error and stop the current func.

pkg/provisioner/controller.go
```diff
@@ -476,14 +490,18 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	// Create OB
 	// Note: do not move ob create/update calls before secret.
 	//   spec.Authentication is lost after create/update, which breaks secret creation
-	ob, err = createObjectBucket(
+	// Note: once at this stage, don't allow ob to be nil if returning with error for cleanup
+	// because ob has important information for the provisioner Delete to clean up resources
+	var createdOB *v1alpha1.ObjectBucket
```
jeffvance, line 495: again, not liking the ob clone here. If `createOB` returned the passed in ob for err!=nil then no info is lost (I think).  What if `createOB` was refactored a bit:
```
func createObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface, retryInterval, retryTimeout time.Duration) (result *v1alpha1.ObjectBucket, err error) {
	logD.Info("creating ObjectBucket", "name", ob.Name)

	err = wait.PollImmediate(retryInterval, retryTimeout, func() (bool, error) {
		result, err = c.ObjectbucketV1alpha1().ObjectBuckets().Create(context.TODO(), ob, metav1.CreateOptions{})
		if errors.IsAlreadyExists(err) {
			err = nil
		} else if err != nil {
			// could be intermittent api error
			log.Error(err, "probably not fatal, retrying")
		}
		return (err == nil), err
	})
	if err == nil {
            return result, nil
       }
      return ob, err
return
}
```

pkg/provisioner/controller.go
BlaineEXE: If updating the obc fails, `obc` becomes `nil` because the fxn returns `nil, <some error>`

pkg/provisioner/controller.go
BlaineEXE: The code that continues is in the deferred cleanup function.

pkg/provisioner/controller.go
BlaineEXE: This is for posterity I guess. It makes it possible for someone to add more code after this that uses the `obc`, and there is less risk that they forget to make sure the `obc` is the latest.

pkg/provisioner/controller.go
```diff
@@ -319,10 +319,13 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	}
 
 	updateOnSuccess := func() error {
```
travisn, line 321: The source of the problem really is embedding the function and allowing it to modify variables in the outer context. This is such a long function anyway, the preferred pattern is to define the function separately and return the computed value(s). Then upon success the caller can assign the variables in its context. But refactoring this would be better saved for another day so we can get the critical fix in today.

pkg/provisioner/controller.go
```diff
@@ -425,14 +409,28 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+	} else if emptyBucket {
 		return fmt.Errorf("provisioner returned empty object bucket")
 	}
 
-	// Fill in known information in the object bucket struct as soon as possible
-	setBasicOBInfo()
+	// This info set on the OB isn't critical for cleanup
 	ob.SetFinalizers([]string{finalizer})
 	ob.Spec.ClaimRef, err = claimRefForKey(key, c.libClientset)
 	if err != nil {
```
BlaineEXE, line 436: I removed the `setBasicOBInfo` function in favor of doing this. This does not change the behavior but I think makes the code more conceptually linear. It does mean the `emptyBucket` check needs to be broken into two parts, and it does mean the error handling for `Provision()/Grant()` comes visually later in the code, but I believe those drawbacks are worth the mental linearity. I feel like this makes code maintenance easier for myself and others.

pkg/provisioner/resourcehandlers.go
```diff
@@ -452,35 +451,44 @@ func deleteObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface) error
...
+	// phase.
+	updateOB := ob.DeepCopy()
+	updateOB.Status.Phase = phase
+
+	result, err = c.ObjectbucketV1alpha1().ObjectBuckets().UpdateStatus(context.TODO(), updateOB, metav1.UpdateOptions{})
+	if err != nil {
+		// return input ob here since result is nil on error returns
+		return ob, fmt.Errorf("failed to update OB %s phase to %q: %v", ob.Name, phase, err)
+	}
+	return result, err
```
BlaineEXE, line 493: I took Jeff's suggestion to make the resource handler functions handle the logic for not allowing ob/obc to become nil. I think this is more clean, but there is some risk in the future if new functions don't follow the same trend. 

At the same time, I took one of Jeff's older suggestions to remove polling. I think this makes the changes I am making here a lot easier to understand conceptually.

pkg/provisioner/controller.go
BlaineEXE: I don't think the secret/cm need to be handled today. They aren't used in the cleanup function, and I don't see any places where they could fall to a `nil` state and cause code issues as it stands today.

pkg/provisioner/controller.go
```diff
@@ -425,14 +409,28 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	} else {
 		ob, err = c.provisioner.Grant(options)
 	}
+	// Record whether the provisioner returned an empty object bucket for error handling use later
+	emptyBucket := (ob == (&v1alpha1.ObjectBucket{}))
```
jeffvance, line 413: Is it possible that a provisioner returns nil for the ob? I think yes and if so it should be handled, but this can be a separate pr.

jeffvance:
`deleteSecretAndWait` (called by `deleteExistingSecretAndConfigMapIfExist`) gets the Secret. But we already have the Secret so I don't see the need to re-get it. This is a nit comment that perhaps can be looked into in a separate PR. Does not impact merge-ability of this pr.

pkg/provisioner/controller.go
BlaineEXE: Sure. I just based this on the previous `else if`

pkg/provisioner/resourcehandlers.go
```diff
@@ -433,7 +432,7 @@ func deleteObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface) error
 
 	logD.Info("removing ObjectBucket finalizer", "name", ob.Name)
 	removeFinalizer(ob)
-	ob, err := c.ObjectbucketV1alpha1().ObjectBuckets().Update(context.TODO(), ob, metav1.UpdateOptions{})
+	_, err := c.ObjectbucketV1alpha1().ObjectBuckets().Update(context.TODO(), ob, metav1.UpdateOptions{})
```
jeffvance, line 435: are you sure there is nothing added to the ob by `Update` that should be returned? I'm thinking of UID, timestamp, revision fields, etc...

BlaineEXE:
> `deleteSecretAndWait` (called by `deleteExistingSecretAndConfigMapIfExist`) gets the Secret. But we already have the Secret so I don't see the need to re-get it. This is a nit comment that perhaps can be looked into in a separate PR. Does not impact merge-ability of this pr.

I did also notice that today. Let's handle that in an update if that sounds good?

pkg/provisioner/controller.go
BlaineEXE: Hmm... there is a compilation issue if I do that. It could be that the check as it was written wasn't working as intended. I can also fix this on follow-up since the check was directly copied from elsewhere and not created new for this PR.

```
cannot compare *ob == (v1alpha1.ObjectBucket literal) (operator == not defined for "github.com/kube-object-storage/lib-bucket-provisioner/pkg/apis/objectbucket.io/v1alpha1".ObjectBucket)
```

BlaineEXE: Created https://github.com/kube-object-storage/lib-bucket-provisioner/issues/203 for this.

pkg/provisioner/resourcehandlers.go
BlaineEXE: Created https://github.com/kube-object-storage/lib-bucket-provisioner/issues/202 for this.

BlaineEXE: Well, the next step is to delete the OB, which is done with `ob.Name` and none of the other ob fields.

pkg/provisioner/resourcehandlers.go
```diff
@@ -452,35 +451,44 @@ func deleteObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface) error
...
 
 func updateClaim(c versioned.Interface, obc *v1alpha1.ObjectBucketClaim, retryInterval, retryTimeout time.Duration) (result *v1alpha1.ObjectBucketClaim, err error) {
-
 	logD.Info("updating", "obc", obc.Namespace+"/"+obc.Name)
-	err = wait.PollImmediate(retryInterval, retryTimeout, func() (bool, error) {
-		result, err = c.ObjectbucketV1alpha1().ObjectBucketClaims(obc.Namespace).Update(context.TODO(), obc, metav1.UpdateOptions{})
-		return (err == nil), err
-	})
-	return
+	result, err = c.ObjectbucketV1alpha1().ObjectBucketClaims(obc.Namespace).Update(context.TODO(), obc, metav1.UpdateOptions{})
```
jeffvance, line 455: should delete retry* args.

pkg/provisioner/resourcehandlers.go
```diff
@@ -452,35 +451,44 @@ func deleteObjectBucket(ob *v1alpha1.ObjectBucket, c versioned.Interface) error
...
+	if err != nil {
+		// return input obc here since result is nil on error returns
+		return obc, fmt.Errorf("failed to update OBC %s/%s: %v", obc.Namespace, obc.Name, err)
+	}
+	return result, err
 }
 
 func updateObjectBucketClaimPhase(c versioned.Interface, obc *v1alpha1.ObjectBucketClaim, phase v1alpha1.ObjectBucketClaimStatusPhase, retryInterval, retryTimeout time.Duration) (result *v1alpha1.ObjectBucketClaim, err error) {
```
jeffvance, line 463: should delete retry* args.

jeffvance, line 463: should delete retry* args.

pkg/provisioner/controller.go
```diff
@@ -425,14 +399,33 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+	if ob != nil {
+		setObjectBucketName(ob, key)
+		ob.Spec.StorageClassName = obc.Spec.StorageClassName
+		if ob.Spec.ReclaimPolicy == nil || *ob.Spec.ReclaimPolicy == corev1.PersistentVolumeReclaimPolicy("") {
+			// Do not blindly overwrite the reclaim policy. The provisioner might have reason to
+			// specify a reclaim policy that is  different from the storage class.
+			ob.Spec.ReclaimPolicy = options.ReclaimPolicy
+		}
+		ob.SetLabels(c.provisionerLabels)
+	}
```
BlaineEXE, line 420: Just had to add an if here to catch cases where the ob returns nil (for a Rook example, this happens if the object store isn't yet ready).

pkg/provisioner/controller.go
```diff
@@ -425,14 +399,33 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+	// Record whether the provisioner returned an empty object bucket for error handling use later
+	emptyBucket := (ob == nil || ob == (&v1alpha1.ObjectBucket{}))
+
+	// Set information about the OB as soon as possible, even before handling errors so that the
+	// deferred cleanup function is not missing critical information needed to clean up.
+	// More note on above: can't check if provisioner returned empty OB after we set OB info below.
+	// The returned OB should only be nil if there is also an error with provisioning; in such a
+	// case, the provisioner should have cleaned up after itself already so we do not need to set
+	// info for OB cleanup.
+	if ob != nil {
```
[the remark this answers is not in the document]
BlaineEXE: I think we could argue it either way. IMO, if the provisioner returns an empty bucket that is non-nil, it probably intended to return without an error, so we should probably still try to clean up (fill in ob details).

pkg/provisioner/controller.go
```diff
@@ -425,14 +399,33 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
...
+			// specify a reclaim policy that is  different from the storage class.
+			ob.Spec.ReclaimPolicy = options.ReclaimPolicy
+		}
+		ob.SetLabels(c.provisionerLabels)
+	}
+
 	if err != nil {
 		return fmt.Errorf("error %s bucket: %v", verb, err)
-	} else if ob == (&v1alpha1.ObjectBucket{}) {
+	} else if emptyBucket {
```
jeffvance, line 424: of course, if the ob != nil then we've updated it above so it's no longer empty but I get the logic...

PR #204, opened 2021-01-05
Remove unnecessary get before secret/cm delete

### LINKED ISSUE #202: Remove unnecessary `Get` calls from `deleteExisingSecretAndConfigMapIfExist`
`deleteSecretAndWait` (called by `deleteExistingSecretAndConfigMapIfExist`) gets the Secret. But we already have the Secret so I don't see the need to re-get it.

BlaineEXE:
Remove unnecessary Get calls before deleting pre-existing
secret/configmap during provisioning.

Fixes #202 

Ping @jeffvance to review without any rush.

### DIFF
--- a/pkg/provisioner/resourcehandlers.go
+++ b/pkg/provisioner/resourcehandlers.go
@@ -235,7 +235,7 @@ func deleteExistingSecretAndConfigMapIfExist(obc *v1alpha1.ObjectBucketClaim, c 
 			return fmt.Errorf("found OBC credentials Secret %q not owned by OBC, assuming this is a user-created resource, user must delete the Secret in order to allow OBC provisioning to continue", secret.Name)
 		}
 		logD.Info("found OBC credentials Secret owned by OBC, deleting and re-creating")
-		err = deleteSecretAndWait(obc, c, defaultRetryBaseInterval, defaultRetryTimeout)
+		err = deleteSecretAndWait(secret, c, defaultRetryBaseInterval, defaultRetryTimeout)
 		if err != nil {
 			return fmt.Errorf("failed to delete existing OBC credentials Secret in order to re-create: %v", err)
 		}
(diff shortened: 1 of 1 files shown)


pkg/provisioner/resourcehandlers.go
```diff
@@ -235,7 +235,7 @@ func deleteExistingSecretAndConfigMapIfExist(obc *v1alpha1.ObjectBucketClaim, c
 			return fmt.Errorf("found OBC credentials Secret %q not owned by OBC, assuming this is a user-created resource, user must delete the Secret in order to allow OBC provisioning to continue", secret.Name)
```
jeffvance, line 235: On [L230](https://github.com/kube-object-storage/lib-bucket-provisioner/pull/204/files#diff-ab88f1a4bb2c3e7d95d10e30594e543aa8649561fde3173455aa2f960f24e967R230) do you want to handle IsNotFound error?

pkg/provisioner/resourcehandlers.go
```diff
@@ -283,28 +283,19 @@ func getExistingConfigMap(obc *v1alpha1.ObjectBucketClaim, c kubernetes.Interfac
...
+	err := releaseSecret(sec, c)
 	if err != nil {
 		return fmt.Errorf("failed to release secret for deletion: %v", err)
 	}
-	err = c.CoreV1().Secrets(obc.Namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
+	err = c.CoreV1().Secrets(sec.Namespace).Delete(context.TODO(), sec.Name, metav1.DeleteOptions{})
 	if err != nil {
 		return err
 	}
 	err = wait.PollImmediate(retryInterval, retryTimeout, func() (done bool, err error) {
```
jeffvance, line 297: why do we need to poll/GET the secret and CM? If the `Delete` works then we know the timestamp is set and we've already removed the finalize, so it's just a matter of (small) time and the secret will be deleted. I can't recall why we should poll here...

jeffvance:
Great that you're cleaning this up! I have just a couple of comments above...

pkg/provisioner/resourcehandlers.go
BlaineEXE: The `getExistingSecret` function handles the `IsNotFound` error by returning a nil secret and nil error.

pkg/provisioner/resourcehandlers.go
BlaineEXE: This function `deleteSecretAndWait` which waits for the secret to be deleted before returning. We need to wait for the secret to be gone before creating the new secret to replace the one being deleted.

jeffvance:
LGTM

PR #207, opened 2021-01-07
Fix broken test and report CI failing correctly

toshipp:
Because the script `hack/go.sh ci-checks` does not report its failure by the exit code, we can not notice a CI failure.
This fixes the issue and also fixes found broken tests.

Signed-off-by: Toshikuni Fukaya <toshikuni-fukaya@cybozu.co.jp>

### DIFF
--- a/hack/go.sh
+++ b/hack/go.sh
@@ -55,9 +55,11 @@ build(){
     cd "${REPO_ROOT}"
     for p in ${PKGS}; do
       echo "go build'ing package $p"
-      go build -a "${p}"
+      go build -a "${p}" || RET=$?
     done
+    return $RET
   )
+  return $?
 }
 
 test(){
@@ -65,9 +67,11 @@ test(){
   (
     cd "${REPO_ROOT}"
     for p in "${PKGS}"; do
-      go test -v "${p}"
+      go test -v "${p}" || RET=$?
     done
+    return $RET
   )
+  return $?
 }
 
 lint(){
@@ -86,8 +90,9 @@ linters(){
 ci-checks(){
     echo "-------- beginning preflight checks"
     lint
-    test
-    build
+    test || RET=$?
+    build || RET=$?
+    return $RET
 }
 
 help(){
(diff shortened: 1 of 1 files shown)


satoru-takeuchi:
LGTM

toshipp:
Could someone review this?

satoru-takeuchi:
Could some one take a look at this PR? It's a small PR and we can't detect some CI failures without applying it.

guymguym:
@jeffvance @copejon ?

hack/go.sh
```diff
@@ -86,8 +90,9 @@ linters(){
 ci-checks(){
     echo "-------- beginning preflight checks"
     lint
-    test
-    build
+    test || RET=$?
```
guymguym, line 93: perhaps safer to define `local RET=0` in all those functions instead of setting a global var?

toshipp: Yeah, It is preferable. I'll fix them

hack/go.sh
```diff
@@ -55,19 +55,23 @@ build(){
     cd "${REPO_ROOT}"
     for p in ${PKGS}; do
       echo "go build'ing package $p"
-      go build -a "${p}"
+      go build -a "${p}" || RET=$?
     done
+    return $RET
   )
+  return $?
```
jeffvance, line 62: It looks like the goal here is to try to build every pkg, even when 1 or more fails. But the err returned is for the last failure only. Perhaps we should fail on the 1st build error, or return a list of error strings of a format like: "$p: $?". What do you think?

toshipp: My intention is to report the build status to CI.
If use fail-fast manner, it is sufficient for the purpose, however, it breaks current behavior which tries to build all packages.
I don't want to change the behavior because some tools may depend on such behavior.
And I personally prefer that all build error is reported.

For 2nd suggestion, it is not effective, IMO.
The script already does echo for package names, and Go also reports which package fails.

Issue #206, opened 2021-01-06
When OBC is deployed using ArgoCD, the reconciliation process does not end


llamerada-jp:
We are trying to deploy Rook's OBC using ArgoCD. We expect that the OB will be created and the status of the OBC will be changed to `Bound`. However, the process does not end with a Pending status for the following reasons.
I'm going to create a patch to fix the problem.

1. ArgoCD deploys an OBC resource with the label `app.kubernetes.io/instance: <ArgoCD's app name>`.
2. Reconciling process of `lib-bucket-provisioner` runs. In the reconciling process, update the label with the `setOBCMetaFields` function.
  When updating a label, existing labels are ignored and only the `bucket-provisioner` label is given.
  https://github.com/kube-object-storage/lib-bucket-provisioner/blob/2e9d6aa76d5895c0b497758e250e9ba69e6c18c1/pkg/provisioner/controller.go#L616
  Also, the `setOBCMetaFields` function discards the return value of the `Update` function.
  https://github.com/kube-object-storage/lib-bucket-provisioner/blob/2e9d6aa76d5895c0b497758e250e9ba69e6c18c1/pkg/provisioner/controller.go#L624
3. Trying to update `spec.bucketName` and `spec.objectBucketName` in OBC after updating OB.
  However, since the OBC is updated by the `setOBCMetaFields` function, it will output the following message and terminate with an error.
  `the object has been modified; please apply your changes to the latest version and try again, requeuing`
  https://github.com/kube-object-storage/lib-bucket-provisioner/blob/2e9d6aa76d5895c0b497758e250e9ba69e6c18c1/pkg/provisioner/controller.go#L331
4. ArgoCD will add `app.kubernetes.io/instance:` label to the OBC again.
5. Repeat from 2.

**How to reproduce it:**

You can reproduce the ArgoCD process by using a shell script like the following.

```
while true
do
  kubectl label objectbucketclaim -n sandbox obc app.kubernetes.io/instance=sandbox 
done
```

```
apiVersion: objectbucket.io/v1alpha1
kind: ObjectBucketClaim
metadata:
  name: obc
  namespace: sandbox
spec:
  generateBucketName: obc-poc
  storageClassName: ceph-hdd-bucket
```

**Environment:**
Kubernetes: 1.18.9
Rook: 1.5.4
Ceph: 15.2.7

toshipp:
This is fixed by https://github.com/rook/rook/pull/7086

PR #208, opened 2021-01-12
fix updating labels and finalizers

llamerada-jp:
fix issue https://github.com/kube-object-storage/lib-bucket-provisioner/issues/206

- merge and set labels and finalizer to OBC

Signed-off-by: Toshikuni Fukaya <toshikuni-fukaya@cybozu.co.jp>
Signed-off-by: H.Muraoka <hiroshi-muraoka@cybozu.co.jp>

### DIFF
--- a/pkg/provisioner/controller.go
+++ b/pkg/provisioner/controller.go
@@ -280,7 +280,7 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	)
 
 	// set finalizer in OBC so that resources cleaned up is controlled when the obc is deleted
-	if err = c.setOBCMetaFields(obc); err != nil {
+	if obc, err = c.setOBCMetaFields(obc); err != nil {
 		return err
 	}
 
@@ -419,7 +419,7 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 			// specify a reclaim policy that is  different from the storage class.
 			ob.Spec.ReclaimPolicy = options.ReclaimPolicy
 		}
-		ob.SetLabels(c.provisionerLabels)
+		addLabels(ob, c.provisionerLabels)
 	}
 
 	if err != nil {
@@ -429,7 +429,7 @@ func (c *obcController) handleProvisionClaim(key string, obc *v1alpha1.ObjectBuc
 	}
 
 	// This info set on the OB isn't critical for cleanup
-	ob.SetFinalizers([]string{finalizer})
+	addFinalizers(ob, []string{finalizer})
 	ob.Spec.ClaimRef, err = claimRefForKey(key, c.libClientset)
 	if err != nil {
 		return fmt.Errorf("error getting reference to OBC: %v", err)
@@ -606,25 +606,24 @@ func (c *obcController) deleteResources(ob *v1alpha1.ObjectBucket, cm *corev1.Co
 }
 
 // Add finalizer and labels to the OBC.
-func (c *obcController) setOBCMetaFields(obc *v1alpha1.ObjectBucketClaim) (err error) {
+func (c *obcController) setOBCMetaFields(obc *v1alpha1.ObjectBucketClaim) (*v1alpha1.ObjectBucketClaim, error) {
 	clib := c.libClientset
 
-	logD.Info("getting OBC to set metadata fields")
-	obc, err = clib.ObjectbucketV1alpha1().ObjectBucketClaims(obc.Namespace).Get(context.TODO(), obc.Name, metav1.GetOptions{})
-	if err != nil {
-		return fmt.Errorf("error getting obc: %v", err)
-	}
+	// Do not make changes directly to the obc used as input. If the update fails, we should return
+	// the obc given as input as it was given so code that comes after can't assume obc is at the
+	// new phase.
+	updateOBC := obc.DeepCopy()
 
-	obc.SetFinalizers([]string{finalizer})
-	obc.SetLabels(c.provisionerLabels)
+	addFinalizers(updateOBC, []string{finalizer})
+	addLabels(updateOBC, c.provisionerLabels)
 
 	logD.Info("updating OBC metadata")
-	obc, err = updateClaim(clib, obc)
+	obcUpdated, err := updateClaim(clib, updateOBC)
 	if err != nil {
-		return fmt.Errorf("error configuring obc metadata: %v", err)
+		return obc, fmt.Errorf("error configuring obc metadata: %v", err)
 	}
 
-	return nil
+	return obcUpdated, nil
 }
 
 func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBucket, error) {
--- a/pkg/provisioner/helpers.go
+++ b/pkg/provisioner/helpers.go
@@ -236,6 +236,35 @@ func storageClassForObjectBucket(ob *v1alpha1.ObjectBucket, c kubernetes.Interfa
 	return class, nil
 }
 
+func addLabels(obj metav1.Object, newLabels map[string]string) {
+	labels := obj.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	for k, v1 := range newLabels {
+		if v2, ok := labels[k]; ok && v1 != v2 {
+			log.Info("key already exists in labels", k, v2, "overwritten by", v1)
+		}
+		labels[k] = v1
+	}
+	obj.SetLabels(labels)
+}
+
+func addFinalizers(obj metav1.Object, newFilalizers []string) {
+	finalizers := obj.GetFinalizers()
+	finalizerMap := make(map[string]struct{})
+	for _, f := range finalizers {
+		finalizerMap[f] = struct{}{}
+	}
+
+	for _, f := range newFilalizers {
+		if _, ok := finalizerMap[f]; !ok {
+			finalizers = append(finalizers, f)
+		}
+	}
+	obj.SetFinalizers(finalizers)
+}
+
 func removeFinalizer(obj metav1.Object) {
 	finalizers := obj.GetFinalizers()
 	for i, f := range finalizers {
--- a/pkg/provisioner/helpers_test.go
+++ b/pkg/provisioner/helpers_test.go
@@ -158,7 +158,7 @@ func Test_objectBucketClaimReconciler_claimFromKey(t *testing.T) {
 		var err error
 
 		if tt.want != nil {
-			if _, err = ec.ObjectbucketV1alpha1().ObjectBucketClaims(tt.want.Namespace).Create(tt.want); err != nil {
+			if _, err = ec.ObjectbucketV1alpha1().ObjectBucketClaims(tt.want.Namespace).Create(context.TODO(), tt.want, metav1.CreateOptions{}); err != nil {
 				t.Errorf("error precreating object: %v", err)
 			}
 		}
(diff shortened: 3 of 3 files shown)


llamerada-jp:
@BlaineEXE 

This PR is ready for review. Would you assign reviewers?

pkg/provisioner/controller.go
```diff
@@ -606,25 +606,25 @@ func (c *obcController) deleteResources(ob *v1alpha1.ObjectBucket, cm *corev1.Co
...
-	obc.SetLabels(c.provisionerLabels)
+	addFinalizers(obcFetched, []string{finalizer})
+	addLabels(obcFetched, c.provisionerLabels)
 
 	logD.Info("updating OBC metadata")
-	obc, err = updateClaim(clib, obc)
+	obcUpdated, err := updateClaim(clib, obcFetched)
 	if err != nil {
-		return fmt.Errorf("error configuring obc metadata: %v", err)
+		return nil, fmt.Errorf("error configuring obc metadata: %v", err)
```
BlaineEXE, line 624: Probably good to return the original OBC in case code that follows needs to process it in an error state  or in case the code in a future time attempts to continue on after an error.

```suggestion
		return obc, fmt.Errorf("error configuring obc metadata: %v", err)
```

pkg/provisioner/controller.go
```diff
@@ -606,25 +606,25 @@ func (c *obcController) deleteResources(ob *v1alpha1.ObjectBucket, cm *corev1.Co
...
-func (c *obcController) setOBCMetaFields(obc *v1alpha1.ObjectBucketClaim) (err error) {
+func (c *obcController) setOBCMetaFields(obc *v1alpha1.ObjectBucketClaim) (*v1alpha1.ObjectBucketClaim, error) {
 	clib := c.libClientset
 
 	logD.Info("getting OBC to set metadata fields")
-	obc, err = clib.ObjectbucketV1alpha1().ObjectBucketClaims(obc.Namespace).Get(context.TODO(), obc.Name, metav1.GetOptions{})
+	obcFetched, err := clib.ObjectbucketV1alpha1().ObjectBucketClaims(obc.Namespace).Get(context.TODO(), obc.Name, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("error getting obc: %v", err)
+		return nil, fmt.Errorf("error getting obc: %v", err)
```
BlaineEXE, line 615: Even in an error, it could be good to return the original OBC itself, just in case code that follows needs to process it in an error state or in case the code in a future time attempts to continue on after an error.

```suggestion
		return obc, fmt.Errorf("error getting obc: %v", err)
```

BlaineEXE (changes requested):

BlaineEXE:
I would like to see a commit body that explains what issue is fixed by the commit. Essentially, that labels and finalizers applied by the user to the OBC are overwritten by the lib.

pkg/provisioner/helpers.go
```diff
@@ -236,6 +236,35 @@ func storageClassForObjectBucket(ob *v1alpha1.ObjectBucket, c kubernetes.Interfa
...
 }
 
+func addLabels(obj metav1.Object, newLabels map[string]string) {
+	labels := obj.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	for k, v1 := range newLabels {
+		if v2, ok := labels[k]; ok {
+			log.Info("key already exists in labels", k, v2)
```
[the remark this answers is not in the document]
llamerada-jp: Fixed to output the new value.

pkg/provisioner/controller.go
llamerada-jp: I will follow the existing error handling way to make the code style consistent, but if we wanna refer to the original obc value out of the invoked function, it's better to use the input value that was given to the function, IMO.
Besides, we might be confused if we go on using a value that is returned together with an error and if we wanna do that, we just should separate the function at the point the function returns a meaningful value.

pkg/provisioner/controller.go
```diff
@@ -606,25 +606,25 @@ func (c *obcController) deleteResources(ob *v1alpha1.ObjectBucket, cm *corev1.Co
 }
 
 // Add finalizer and labels to the OBC.
-func (c *obcController) setOBCMetaFields(obc *v1alpha1.ObjectBucketClaim) (err error) {
+func (c *obcController) setOBCMetaFields(obc *v1alpha1.ObjectBucketClaim) (*v1alpha1.ObjectBucketClaim, error) {
 	clib := c.libClientset
 
 	logD.Info("getting OBC to set metadata fields")
-	obc, err = clib.ObjectbucketV1alpha1().ObjectBucketClaims(obc.Namespace).Get(context.TODO(), obc.Name, metav1.GetOptions{})
+	obcFetched, err := clib.ObjectbucketV1alpha1().ObjectBucketClaims(obc.Namespace).Get(context.TODO(), obc.Name, metav1.GetOptions{})
```
jeffvance, line 613: 1) why does the OBC need to be retrieved again here? Nothing has changed has it (other than perhaps a window where the user changes it right after creating it)?
2) Why can't we append the lib label to the existing label(s) in the OBC?

pkg/provisioner/helpers.go
```diff
@@ -236,6 +236,35 @@ func storageClassForObjectBucket(ob *v1alpha1.ObjectBucket, c kubernetes.Interfa
 	return class, nil
 }
 
+func addLabels(obj metav1.Object, newLabels map[string]string) {
```
jeffvance, line 239: sorry I didn't look it up, but is this a pointer or by-value?

pkg/provisioner/helpers.go
```diff
@@ -236,6 +236,35 @@ func storageClassForObjectBucket(ob *v1alpha1.ObjectBucket, c kubernetes.Interfa
...
+}
+
+func addFinalizers(obj metav1.Object, newFilalizers []string) {
+	finalizers := obj.GetFinalizers()
+	finalizerMap := make(map[string]struct{})
+	for _, f := range finalizers {
+		finalizerMap[f] = struct{}{}
+	}
+
+	for _, f := range newFilalizers {
```
jeffvance, line 260: why does the lib override (with a warning) user labels that conflict with ours, but we don't override user finalizers that conflict?

jeffvance: I'm not sure why the `finalizerMap` is needed? Can't we append the lib finalzers array to the existing finalizers array returned by `getFinalizers`?

pkg/provisioner/controller.go
BlaineEXE: I had a similar question. Existing code was retrieving the OBC again. Do we need to do this in case the OBC gets updated twice in rapid succession by the user before this reconcile happens?

jeffvance: @BlaineEXE To answer your question about an OBC that's created -> lib GETs it -> user edits it -> lib's copy is stale:  do we still ignore update events on the OBC? If yes then this window exists and it seems that we would lose the user's changes. But, even with the GET here we still have the same window, just smaller, right?

BlaineEXE: Yes, I think that's right. If the user updates twice rapidly, then a second reconcile will still be queued to happen again after this reconcile finishes IIUC. So this isn't necessary. I think @jeffvance and I are agreed that this `Get`, despite being in the repo pre-existing, should be removed. 

@llamerada-jp you can look at some of the other functions to see how update attempts use `DeepCopy`-ed objects to not modify the object used as input `obc` in this case.

pkg/provisioner/helpers.go
BlaineEXE: I think the lib should override conflicting finalizers also.

pkg/provisioner/helpers.go
BlaineEXE: `metav1.Object` is an interface. I did a quick test in the go playground to make sure for myself that this will work, and I think it works as expected.

https://play.golang.org/p/7EtZ-WiZSJT

pkg/provisioner/controller.go
BlaineEXE: I don't think there is anything wrong with returning a value under an error condition in this case. This pattern is used elsewhere in the resourcehandlers. Consider pseudocode below where cleanUp() might need info from value.
```
value = some starting value
value, err = mutateValue(value)
if err {
  cleanUp(value)
  return err
}

```

BlaineEXE (changes requested):
Please fix the commit to reduce the subject line length to 54 chars or less. Also include a body that explains what this fix does and why it's needed. Limit body line length to 74 chars. Example...

```
allow user labels and finalizers on obcs

What does this fix do?
Mention that user-defined labels/finalizers may be
overridden if they conflict with the labels/finalizers the
OBC want's to add.

Why is this fix needed? Which users benefit?

Signed-off-by: blah <blah>
```

pkg/provisioner/controller.go
jeffvance: I tend to agree because, in general, code check for `err != nil` before it processes other func returns...

pkg/provisioner/helpers.go
llamerada-jp: If we simply add a finalizer, the value will be duplicated if retries happen.  For this reason, we append only when there is no finalizer in the list.

jeffvance: Good point about dealing with retries and I agree that finalizers map supports this. Since labels are a map and finalizers are an array, the logic will not be symmetric between the two funcs. However, won't retries spam the log for adding labels when the key already exists from try N-1?

llamerada-jp: Indeed. How about logging only if label value is changed.

pkg/provisioner/controller.go
llamerada-jp: I update it using DeepCopy.

BlaineEXE (changes requested):
Sorry to forget this for a few days. My work laptop broke, and I lost track of some things.

Also, please revise the commit message. Do not put my suggested statements/questions into the message. Just make sure that the information I suggested is explained in your words.

Thanks!

pkg/provisioner/helpers.go
```diff
@@ -236,6 +236,35 @@ func storageClassForObjectBucket(ob *v1alpha1.ObjectBucket, c kubernetes.Interfa
 	return class, nil
 }
 
+func addLabels(obj metav1.Object, newLabels map[string]string) {
+	labels := obj.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	for k, v1 := range newLabels {
+		if v2, ok := labels[k]; ok && v1 != v2 {
```
jeffvance, line 245: Good way not to spam log on retries!

pkg/provisioner/helpers.go
```diff
@@ -236,6 +236,35 @@ func storageClassForObjectBucket(ob *v1alpha1.ObjectBucket, c kubernetes.Interfa
...
+			log.Info("key already exists in labels", k, v2, "overwritten by", v1)
+		}
+		labels[k] = v1
+	}
+	obj.SetLabels(labels)
+}
+
+func addFinalizers(obj metav1.Object, newFilalizers []string) {
```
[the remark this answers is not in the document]
llamerada-jp: We added some tests for `addFinalizers`.

BlaineEXE (approved):

BlaineEXE:
@llamerada-jp now that this is merged, the next step is to include these changes into Rook by updating the go module reference.

toshipp:
@BlaineEXE 
Thanks!
I am a colleague of @llamerada-jp.
We will send a PR to Rook project.

PR #210, opened 2021-03-11
api: add kubebuilder marker for scope

leseb:
Kubebuiler will apply the "Namespaced" scope by default if not specified
otherwise. The ObjectBucket resource is Cluster scoped so we need to
force that.


dannyzaken:
LGTM

PR #212, opened 2021-03-11
api: fix resource marker for kubebuilder

leseb:
We expect all the resource to be on the same marker line.


BlaineEXE (approved):

BlaineEXE:
Doesn't objectbucketclaim_types.go also need updated?

leseb:
> objectbucketclaim_types

It has a single parameter to the resource so it's fine :)

PR #213, opened 2021-03-24
api: add api support for Update events

### LINKED ISSUE #195: Updating of additonal config field of OBC
I am not sure whether bug or current design limitation. In rook quota for OBC is supported with help of `additionalConfig` string Map. When I try to update the field it is not working via `kubectl edit` or `kubectl apply` the request is not handled properly.

I can see following logs from rook-operator
```console
I1014 10:05:53.347471       6 controller.go:210]  "msg"="reconciling claim" "key"="default/ceph-delete-bucket" 
I1014 10:05:53.347535       6 helpers.go:80]  "msg"="getting claim for key" "key"="default/ceph-delete-bucket" 
I1014 10:05:53.349255       6 helpers.go:181]  "msg"="getting ObjectBucketClaim's StorageClass" "key"="default/ceph-delete-bucket" 
I1014 10:05:53.353931       6 helpers.go:186]  "msg"="got StorageClass" "key"="default/ceph-delete-bucket" "name"="rook-ceph-delete-bucket"
I1014 10:05:53.354343       6 helpers.go:63]  "msg"="checking OBC for OB name, this indicates provisioning is complete" "key"="default/ceph-delete-bucket" "ceph-delete-bucket"=null
I1014 10:05:53.354404       6 helpers.go:65]  "msg"="provisioning already completed" "key"="default/ceph-delete-bucket" "ObjectBucket"="obc-default-ceph-delete-bucket"
I1014 10:05:53.354424       6 controller.go:246]  "msg"="skipping provision" "key"="default/ceph-delete-bucket"
```

If I understand correctly current code follow in `lib-bucket-provisioner` the request reaches `syncHandler`. 

```go
   // ***********************
    // Delete or Revoke Bucket
    // ***********************
    if obc.ObjectMeta.DeletionTimestamp != nil {
        log.Info("OBC deleted, proceeding with cleanup")
        return c.handleDeleteClaim(key, obc)
    }

    // *******************************************************
    // Provision New Bucket or Grant Access to Existing Bucket
    // *******************************************************
    if !shouldProvision(obc) {
        log.Info("skipping provision")
        return nil
    }

    // update the OBC's status to pending before any provisioning related errors can occur
    obc, err = updateObjectBucketClaimPhase(
        c.libClientset,
```
Firsts it check for whether request delete/revokes, if not checks whether provisioning is needed via `shouldProvision`. The `shouldProvision` checks whether `OB` name is present on `OBC` CRD. If exists it returns true. In case of updating `OBC` field this always exist and skips further part of code. Is this intended behavior? 

CCing @yuvalif

thotz:
Adding support to handle Update Events of OBC, currently `AdditionalConfig` needs this change.
Other fields are immutable
Fixes #195

### DIFF
--- a/pkg/provisioner/api/provisioner.go
+++ b/pkg/provisioner/api/provisioner.go
@@ -43,6 +43,8 @@ type Provisioner interface {
 	Provision(options *BucketOptions) (*v1alpha1.ObjectBucket, error)
 	// Grant should be implemented to handle access to existing buckets
 	Grant(options *BucketOptions) (*v1alpha1.ObjectBucket, error)
+	// Update should be implemented to handle cases when additional config are modified
+	Update(ob *v1alpha1.ObjectBucket) error
 	// Delete should be implemented to handle bucket deletion
 	Delete(ob *v1alpha1.ObjectBucket) error
 	// Revoke should be implemented to handle removing bucket access
--- a/pkg/provisioner/controller.go
+++ b/pkg/provisioner/controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"reflect"
 	"strconv"
 	"time"
 
@@ -99,6 +100,11 @@ func NewController(provisionerName string, provisioner api.Provisioner, clientse
 			if newObc.ObjectMeta.DeletionTimestamp != nil && oldObc.ObjectMeta.DeletionTimestamp != nil {
 				return
 			}
+
+			if !updateSupported(oldObc, newObc) {
+				return
+			}
+
 			// handle this update
 			ctrl.enqueueOBC(new)
 		},
@@ -117,7 +123,7 @@ func (c *obcController) Start(stopCh <-chan struct{}) error {
 	defer c.queue.ShutDown()
 
 	if !cache.WaitForCacheSync(stopCh, c.obcHasSynced, c.obHasSynced) {
-		return fmt.Errorf("failed to waith for caches to sync ")
+		return fmt.Errorf("failed to wait for caches to sync ")
 	}
 	count := 1
 	if threadiness, set := os.LookupEnv("LIB_BUCKET_PROVISIONER_THREADS"); set {
@@ -241,13 +247,17 @@ func (c *obcController) syncHandler(key string) error {
 		return c.handleDeleteClaim(key, obc)
 	}
 
+	// ***********************
+	// Update Events on Bucket
+	// ***********************
+	if !shouldProvision(obc) {
+		log.Info("updating OBC")
+		return c.handleUpdateClaim(key, obc, class)
+	}
+
 	// *******************************************************
 	// Provision New Bucket or Grant Access to Existing Bucket
 	// *******************************************************
-	if !shouldProvision(obc) {
-		log.Info("skipping provision")
-		return nil
-	}
 
 	// update the OBC's status to pending before any provisioning related errors can occur
 	obc, err = updateObjectBucketClaimPhase(
@@ -637,4 +647,63 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
 		return nil, err
 	}
 	return ob, nil
+}
+
+func updateSupported(old, new *v1alpha1.ObjectBucketClaim) bool {
+	// The only field supported for update is obc.spec.additionalConfig
+	if reflect.DeepEqual(new.Spec, old.Spec) {
+		return false
+	}
+	// create copy of old spec, and set the new spec's additionalConfig on it
+	oldspec := old.Spec.DeepCopy()
+	overwriteAdditionalConfig(new.Spec.AdditionalConfig, oldspec.AdditionalConfig)
+	if !reflect.DeepEqual(*oldspec, new.Spec) {
+		// new OBC spec has changed something other than additionalConfig
+		log.Error(nil, "invalid changes to OBC. only additionalConfig can be updated")
+		return false
+	}
+	return true
+}
+
+func (c *obcController) handleUpdateClaim(key string, obc *v1alpha1.ObjectBucketClaim, class *storagev1.StorageClass) error {
+	ob, err := getObjectBucketFromClaimKey(key, obc, c.libClientset, defaultRetryBaseInterval, defaultRetryTimeout)
+	if err != nil {
+		log.Error(err, "getting ob failed", "key", key)
+		return err
+	}
+
+	log.Info("Additional config changed, hence updating OB")
+	tmp := make(map[string]string)
+
+	overwriteAdditionalConfig(ob.Spec.Endpoint.AdditionalConfigData, tmp)
+	overwriteAdditionalConfig(obc.Spec.AdditionalConfig, ob.Spec.Endpoint.AdditionalConfigData)
+	err = c.provisioner.Update(ob)
+	if err != nil {
+		log.Error(err, "updating from Provisioner failed")
+		return err
+	}
+
+	err = wait.PollImmediate(defaultRetryTimeout, defaultRetryBaseInterval, func() (done bool, err error) {
+		_, err = updateObjectBucket(ob, c.libClientset)
+		if err != nil {
+			log.Error(err, "updating OB failed, retrying...")
+			return false, err
+		}
+		return true, nil
+	})
+	if err != nil {
+		log.Error(err, "updating OB failed, reverting provisioner to original value")
+		overwriteAdditionalConfig(tmp, ob.Spec.Endpoint.AdditionalConfigData)
+		err = c.provisioner.Update(ob)
+	}
+	return err
+}
+
+func overwriteAdditionalConfig(srcAdditionalConfig, destAdditionalConfig map[string]string) {
+	for key, _ := range destAdditionalConfig {
+		delete(destAdditionalConfig, key)
+	}
+	for key, value := range srcAdditionalConfig {
+		destAdditionalConfig[key] = value
+	}
 }
--- a/pkg/provisioner/fakeinterface_test.go
+++ b/pkg/provisioner/fakeinterface_test.go
@@ -59,3 +59,11 @@ func (p *fakeProvisioner) Revoke(ob *v1alpha1.ObjectBucket) (err error) {
 	}
 	return err
 }
+
+// Update provides a simple method for testing purposes
+func (p *fakeProvisioner) Update(ob *v1alpha1.ObjectBucket) (err error) {
+	if ob == nil {
+		err = fmt.Errorf("got nil object bucket pointer")
+	}
+	return err
+}


thotz:
@BlaineEXE @jeffvance @copejon @leseb PTAL

pkg/provisioner/api/provisioner.go
```diff
@@ -43,6 +43,8 @@ type Provisioner interface {
 	Provision(options *BucketOptions) (*v1alpha1.ObjectBucket, error)
 	// Grant should be implemented to handle access to existing buckets
 	Grant(options *BucketOptions) (*v1alpha1.ObjectBucket, error)
+	// Update should be implemented to handle updation of existing buckets
+	Update(options *BucketOptions) error
```
jeffvance, line 47: I think this is a breaking change in that **all** provisioners will need to add an Update method when they recompile.

pkg/provisioner/controller.go
```diff
@@ -245,8 +246,8 @@ func (c *obcController) syncHandler(key string) error {
 	// Provision New Bucket or Grant Access to Existing Bucket
 	// *******************************************************
 	if !shouldProvision(obc) {
-		log.Info("skipping provision")
-		return nil
+		log.Info("updating OBC")
```
jeffvance, line 249: comments on lines 238 and 246 should include that the event could also be update.

pkg/provisioner/controller.go
```diff
@@ -638,3 +639,38 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
 }
+
+func (c *obcController) handleUpdateClaim(key string, obc *v1alpha1.ObjectBucketClaim, class *storagev1.StorageClass) error {
+	ob, err := getObjectBucketFromClaimKey(key, obc, c.libClientset, defaultRetryBaseInterval, defaultRetryTimeout)
+	if err != nil {
+		log.Error(err, "getting ob failed", "key", key)
+		return err
+	}
+	if reflect.DeepEqual(ob.Spec.Endpoint.AdditionalConfigData, obc.Spec.AdditionalConfig) {
+		logD.Info("No updation required")
```
jeffvance, line 650: "updation" is not a word (in English). Try "Only additionalConfigData" can be changed. Update ignored."

pkg/provisioner/controller.go
```diff
@@ -638,3 +639,38 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+		return nil
+	} else {
+		logD.Info("Additional config changed, OB updation required")
```
jeffvance, line 653: "updation" is not a word.

pkg/provisioner/controller.go
```diff
@@ -638,3 +639,38 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+		bucketName, _ := composeBucketName(obc)
+		options := &api.BucketOptions{
+			ReclaimPolicy:     class.ReclaimPolicy,
```
jeffvance, line 656: The storage class could have been deleted and then recreated with the same name. We don't want to potentially change the retainPolicy. Why are any field being set other than additionalConfigData?

pkg/provisioner/controller.go
```diff
@@ -638,3 +639,38 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+	if err != nil {
+		log.Error(err, "getting ob failed", "key", key)
+		return err
+	}
+	if reflect.DeepEqual(ob.Spec.Endpoint.AdditionalConfigData, obc.Spec.AdditionalConfig) {
+		logD.Info("No updation required")
+		return nil
+	} else {
+		logD.Info("Additional config changed, OB updation required")
+		bucketName, _ := composeBucketName(obc)
```
jeffvance, line 654: why do we need to re-compose the bucketName?

pkg/provisioner/controller.go
```diff
@@ -638,3 +639,38 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+		options := &api.BucketOptions{
+			ReclaimPolicy:     class.ReclaimPolicy,
+			BucketName:        bucketName,
+			ObjectBucketClaim: obc.DeepCopy(),
+			Parameters:        class.Parameters,
+		}
+		err = c.provisioner.Update(options)
+		if err != nil {
+			log.Error(err, "updation from Provisioner failed")
```
jeffvance, line 663: "updation"

pkg/provisioner/controller.go
```diff
@@ -638,3 +639,38 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+			return err
+		}
+		for k, v := range obc.Spec.AdditionalConfig {
+			ob.Spec.Endpoint.AdditionalConfigData[k] = v
+		}
+		ob, err = updateObjectBucket(ob, c.libClientset)
+		if err != nil {
+			log.Error(err, "OB updation failed")
```
jeffvance, line 671: "updation"

jeffvance:
There is no error reported if the user tries to update OBC fields other than additionalCongifData.
Also, have you consider additionalConfigData other than quota (which the user is allowed to change). For example, the storage class's parameters field can contain bucket config data and is passed to the provisioner. What happens if the user changes this config data via an updating their obc's additionalConfigData. It seems now that a user is allowed to change _any_ bucket config data.

BlaineEXE:
Are these changes even needed any more given this comment? https://github.com/kube-object-storage/lib-bucket-provisioner/issues/195#issuecomment-742282150

thotz:
> Are these changes even needed any more given this comment? [#195 (comment)](https://github.com/kube-object-storage/lib-bucket-provisioner/issues/195#issuecomment-742282150)

Atleast at that time focus was needed for bucket notifications(decided to go with `label notations` instead of `additionalconfig`), now users have similar request for quota as well https://github.com/rook/rook/issues/7146

pkg/provisioner/api/provisioner.go
thotz: That happens only if they update the latest version of the library right? Can we release note or something in which we can mention about this change

pkg/provisioner/controller.go
thotz: I will pass `OBC` instead of options that make more sense, provisioner need to ob info (backend bucket and backend user)

thotz:
> There is no error reported if the user tries to update OBC fields other than additionalCongifData.
> Also, have you consider that the user can mutate additionalConfigData other than quota (which the user is allowed to change). For example, the storage class's parameters field can contain bucket config data and is passed to the provisioner. What happens if the user changes this config data by updating their obc's additionalConfigData. It seems now a user is allowed to change _any_ bucket config data.

@jeffvance : I agree that other fields apart from `additionalConfigData` should not be modified, this field is opaque to the library so it won't able to know changes made to `additionalConfigData`. Sorry I didn't fully understand how `storageclass.Parameter` related to `additionalConfigData`

jeffvance:
> @jeffvance : I agree that other fields apart from `additionalConfigData` should not be modified, this field is opaque to the library so it won't able to know changes made to `additionalConfigData`. Sorry I didn't fully understand how `storageclass.Parameter` related to `additionalConfigData`

1) what happens if the user modifies OBC fields other than additionalConfigData? It looks like there will be no error reported.
2) Re. storage class and parameters- nevermind. I didn't remember that your are passing an options struct  to the Update method, which define the field the Update method sees.  Be careful if change to pass a map to Update because then there's a possibility of collisions with the StorageClass.parameters config map.

pkg/provisioner/api/provisioner.go
jeffvance: Agree

pkg/provisioner/controller.go
jeffvance: What is `additionalState`?

thotz: I guess I will pass OB than separate parameters, so this API will resemble to Delete() or Grant(). `Additional State` another string map that stores information created by the backend of provisioner like user/owner of the bucket
```console
Spec:
  Additional State:
    Ceph User:  ceph-user-up2Dphkq
```

jeffvance:
> now users have similar request for quota as well rook/rook#7146
So do we want to also support notification via additionalConfigData?

jeffvance:
From above:
> what happens if the user modifies OBC fields other than additionalConfigData? It looks like there will be no error reported.

This has not been addressed.

@BlaineEXE what are your thoughts?

thotz:
> > at that time focus was needed for bucket notifications (decided to go with label notations instead of additionalconfig)
> 
> So do we want to also support notification via additionalConfigData?

Currently not planned to do so

pkg/provisioner/controller.go
```diff
@@ -99,6 +100,12 @@ func NewController(provisionerName string, provisioner api.Provisioner, clientse
 			if newObc.ObjectMeta.DeletionTimestamp != nil && oldObc.ObjectMeta.DeletionTimestamp != nil {
 				return
 			}
+			if !reflect.DeepEqual(newObc.Spec, oldObc.Spec) {
+				if reflect.DeepEqual(newObc.Spec.AdditionalConfig, oldObc.Spec.AdditionalConfig) {
+					log.Error(nil, "Only Additional Config can be modified for an OBC")
```
thotz, line 105: @jeffvance : will this suffice as an error check for modifying other fields apart from `additionalconfig`, so that update happens specifically.

jeffvance: The `if` in [L100](https://github.com/kube-object-storage/lib-bucket-provisioner/pull/213/files#diff-1e346e3430c7a1bec9a1fde398367f33b921671c6c3ced081ad42e4d8e345522R100) is checking that if _both_ the old and new OBCs are being deleted then `return`.
I think you only need to check if the new OBC is being deleted and then return.

jeffvance: > will this suffice as an error check for modifying other fields apart from additionalconfig?

Yes. I think I missed seeing the first DeepEqual on spec.

BlaineEXE: Go error messages are supposed to not start with capital letters. I also suggest rewording for clarity.

```suggestion
					log.Error(nil, "invalid changes to OBC. only additionalConfig can be updated after creation")
```

pkg/provisioner/controller.go
```diff
@@ -638,3 +649,49 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+		for key, value = range ob.Spec.Endpoint.AdditionalConfigData {
+			tmp[key] = value
+			delete(ob.Spec.Endpoint.AdditionalConfigData, key)
+		}
+
+		for key, value = range obc.Spec.AdditionalConfig {
+			ob.Spec.Endpoint.AdditionalConfigData[key] = value
+		}
+
+		err = c.provisioner.Update(ob)
```
BlaineEXE, line 678: I spent some time a while ago making sure that the `Create` method was idempotent.

What about the option of not creating a new `Update` method that has to be implemented by users (i.e., Rook) and instead call Create again? Or does that also pose different issues?

pkg/provisioner/controller.go
BlaineEXE: I read that this code checks 
```go
if new spec has changed {
  if new additional config is the same as old additional config {
    logError
    return
    // this part seems correct to me
  }
}
```

But what if a user modifies both the spec and the additionalConfig?

I think the check needs to be more like:
```go
// create copy of old spec, and set the new spec's additionalConfig on it
oldSpecWithNewAdditionalConfig := oldObc.Spec.DeepCopy()
oldSpecWithNewAdditinoalConfig.AdditionalConfig = newObc.Spec.AdditionalConfig.DeepCopy()

if !reflect.DeepEqual(newObc.Spec, oldObc.Spec) {
    // new OBC is changed
    if !reflect.DeepEqual(oldSpecWithNewAdditionalConfig, newObc.Spec) {
        // new OBC has changed something other than additionalConfig
        log.Error(nil, "invalid changes to OBC. only additionalConfig can be updated after creation")
        return
    }
}
```

BlaineEXE (changes requested):

jeffvance:
> But what if a user modifies both the spec and the additionalConfig?

Good catch @BlaineEXE !

pkg/provisioner/controller.go
```diff
@@ -638,3 +656,48 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+	}
+	if reflect.DeepEqual(ob.Spec.Endpoint.AdditionalConfigData, obc.Spec.AdditionalConfig) {
+		return nil
+	} else {
+		logD.Info("Additional config changed, hence updating OB")
+		tmp := make(map[string]string)
+
+		copyAdditionalConfig(ob.Spec.Endpoint.AdditionalConfigData, tmp)
+		clearAdditionalConfig(ob.Spec.Endpoint.AdditionalConfigData)
+		copyAdditionalConfig(obc.Spec.AdditionalConfig, ob.Spec.Endpoint.AdditionalConfigData)
```
BlaineEXE, line 674: What if, instead of calling `clear...` then `copy...` (which is more of a merge by how it is implemented) there was a function to `overwriteAdditionalConfig` that did both clear and copy (merge) steps? I think that will make it less likely in the future that someone forgets to call `clear` before `copy` when they should do so.

After further thought, It might be more clear to read the functions at a glance as below, which removes the need for `clearAdditionalConfig()` entirely and makes it clear what is being copied and where it is being copied into:
```go
tmp := copyAdditionalConfig(ob.Spec.Endpoint.AdditionalConfigData) // copy into tmp
// copy into (and overwrite) OB endpoint's additional config
ob.Spec.Endpoint.AdditionalConfigData = copyAdditionalConfig(obc.Spec.AdditionalConfig) 

func copyAdditionalConfig(a map[string]string) map[string]string {
    ret := map[string]string{}
    for k, v := range a {
        ret[k] = v
    }
    return ret
}
```

pkg/provisioner/controller.go
```diff
@@ -638,3 +656,48 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+			clearAdditionalConfig(ob.Spec.Endpoint.AdditionalConfigData)
+			copyAdditionalConfig(tmp, ob.Spec.Endpoint.AdditionalConfigData)
+			log.Error(err, "updating from Provisioner failed")
+			return err
+		}
+		ob, err = updateObjectBucket(ob, c.libClientset)
+		if err != nil {
+			log.Error(err, "updating OB failed")
+			return err
+		}
```
BlaineEXE, line 688: Is it necessary to update the ObjectBucket if the Update fails? 

From how I read the function up until this point, If the update has failed, I think there are no changes being made to the ObjectBucket. If that is the case, we should probably just not update the bucket at all. But maybe I missed something.

pkg/provisioner/controller.go
```diff
@@ -638,3 +656,48 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+		clearAdditionalConfig(ob.Spec.Endpoint.AdditionalConfigData)
+		copyAdditionalConfig(obc.Spec.AdditionalConfig, ob.Spec.Endpoint.AdditionalConfigData)
+
+		err = c.provisioner.Update(ob)
+		if err != nil {
+			// Provisioner returns error, hence reverting values
+			clearAdditionalConfig(ob.Spec.Endpoint.AdditionalConfigData)
+			copyAdditionalConfig(tmp, ob.Spec.Endpoint.AdditionalConfigData)
+			log.Error(err, "updating from Provisioner failed")
+			return err
```
BlaineEXE, line 682: These steps are only necessary if the OB actually needs its resource updated following a provisioner `Update()` failure, right?

pkg/provisioner/controller.go
```diff
@@ -638,3 +656,48 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+
+func (c *obcController) handleUpdateClaim(key string, obc *v1alpha1.ObjectBucketClaim, class *storagev1.StorageClass) error {
+	ob, err := getObjectBucketFromClaimKey(key, obc, c.libClientset, defaultRetryBaseInterval, defaultRetryTimeout)
+	if err != nil {
+		log.Error(err, "getting ob failed", "key", key)
+		return err
+	}
+	if reflect.DeepEqual(ob.Spec.Endpoint.AdditionalConfigData, obc.Spec.AdditionalConfig) {
+		return nil
+	} else {
```
jeffvance, line 668: more "go-like" to omit the `else` since you've returned in the line above.

BlaineEXE: Ah, yes. I think some linters would complain about this too.

pkg/provisioner/controller.go
```diff
@@ -638,3 +656,48 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+		logD.Info("Additional config changed, hence updating OB")
+		tmp := make(map[string]string)
+
+		copyAdditionalConfig(ob.Spec.Endpoint.AdditionalConfigData, tmp)
+		clearAdditionalConfig(ob.Spec.Endpoint.AdditionalConfigData)
+		copyAdditionalConfig(obc.Spec.AdditionalConfig, ob.Spec.Endpoint.AdditionalConfigData)
+
+		err = c.provisioner.Update(ob)
```
jeffvance, line 676: I know an earlier version of this pr passed an options struct to Update and now you're passing the entire OB, with the need to save a tmp copy of additionalConfig, etc. What is the reason for not just passing to Update a single parameter named `additionalConfig` as a map[string]string? Then provisioners don't need to worry about what else may have changed in the OB. It feels safer and simpler, but at the cost of being less future-proof if it's decided to support more OB mutations.

thotz: @jeffvance : If I pass only the `additionalconfig` the provisioned does not much idea about details like which user or bucket(which might be opaque to library and different for each) it is referring to. Easier way to pass `OB` information like `Delete()` or `Revoke()`. Or I need to figure what all must is needed(from my point of view I can only check from Rook)

pkg/provisioner/controller.go
```diff
@@ -638,3 +655,43 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+		log.Error(err, "updating OB failed")
+		// revert to old values
+		overwriteAdditionalConfig(tmp, ob.Spec.Endpoint.AdditionalConfigData)
+		c.provisioner.Update(ob)
+		return err
+	}
+	return nil
+}
+
+func overwriteAdditionalConfig(srcAdditionalConfig map[string]string, destAdditionalConfig map[string]string) {
```
jeffvance, line 690: nit:
`func overwriteAdditionalConfig(srcAdditionalConfig,  destAdditionalConfig map[string]string) {`

pkg/provisioner/controller.go
```diff
@@ -638,3 +655,43 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+		log.Error(err, "getting ob failed", "key", key)
+		return err
+	}
+	if reflect.DeepEqual(ob.Spec.Endpoint.AdditionalConfigData, obc.Spec.AdditionalConfig) {
+		return nil
+	}
+	log.Info("Additional config changed, hence updating OB")
+	tmp := make(map[string]string)
+
+	overwriteAdditionalConfig(ob.Spec.Endpoint.AdditionalConfigData, tmp)
```
jeffvance, line 671: Suggest delete L669 and here replace with:
`tmp := ob.Spec.Endpoint.AdditionalConfigData`

pkg/provisioner/controller.go
```diff
@@ -638,3 +655,43 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+	overwriteAdditionalConfig(obc.Spec.AdditionalConfig, ob.Spec.Endpoint.AdditionalConfigData)
+
+	err = c.provisioner.Update(ob)
+	if err != nil {
+		log.Error(err, "updating from Provisioner failed")
+		return err
+	}
+	ob, err = updateObjectBucket(ob, c.libClientset)
+	if err != nil {
```
jeffvance, line 680: It seems like this is a good place to do a retry inline vs. requeuing on error. It's potentially expensive for the provisioner to update the new additionalConfig, then update back to the old value, then the event is requeued, then provisioner updates to new config again, etc.

pkg/provisioner/controller.go
```diff
@@ -99,6 +100,18 @@ func NewController(provisionerName string, provisioner api.Provisioner, clientse
 			if newObc.ObjectMeta.DeletionTimestamp != nil && oldObc.ObjectMeta.DeletionTimestamp != nil {
 				return
 			}
+
```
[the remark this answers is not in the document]
thotz: `label` and `annotation` is not part of `ObjectBucketClaimSpec` this `Update` API or previous version of `Updatefunc` was not checking that. If needed we can define different API for those ones. And even `Provisioner`(like Rook) can itself(or has authority) to check for labels and annotations. Or is the duty to library inform the `Provisioner` for these changes.

pkg/provisioner/controller.go
```diff
@@ -99,6 +100,18 @@ func NewController(provisionerName string, provisioner api.Provisioner, clientse
...
+			// create copy of old spec, and set the new spec's additionalConfig on it
+			oldSpecWithNewAdditionalConfig := oldObc.Spec.DeepCopy()
+			overwriteAdditionalConfig(newObc.Spec.AdditionalConfig, oldSpecWithNewAdditionalConfig.AdditionalConfig)
+			if !reflect.DeepEqual(newObc.Spec, oldObc.Spec) {
+				// new OBC is changed
+				if !reflect.DeepEqual(*oldSpecWithNewAdditionalConfig, newObc.Spec) {
```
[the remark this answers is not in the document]
thotz: without pointer reference the check was always returning false

pkg/provisioner/controller.go
```diff
@@ -638,3 +655,43 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
 	return ob, nil
 }
+
+func (c *obcController) handleUpdateClaim(key string, obc *v1alpha1.ObjectBucketClaim, class *storagev1.StorageClass) error {
+	ob, err := getObjectBucketFromClaimKey(key, obc, c.libClientset, defaultRetryBaseInterval, defaultRetryTimeout)
+	if err != nil {
+		log.Error(err, "getting ob failed", "key", key)
+		return err
+	}
+	if reflect.DeepEqual(ob.Spec.Endpoint.AdditionalConfigData, obc.Spec.AdditionalConfig) {
```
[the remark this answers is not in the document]
thotz: I guess this check may reductant

pkg/provisioner/controller.go
thotz: are suggesting the following
`for i=0; i<3 &&  err !=nil; i++ {
ob, err = updateObjectBucket(ob, c.libClientset)
}

pkg/provisioner/controller.go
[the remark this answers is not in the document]
jeffvance: Is it always true now? You want to make sure you are not comparing an address with a spec.

jeffvance:
Typo L133: "waith" -> "wait"

pkg/provisioner/controller.go
jeffvance: No. Look at examples of using the apimachinary's _wait_ package; (PollImmediate)[https://github.com/kubernetes/apimachinery/blob/v0.21.0/pkg/util/wait/wait.go#L440]. Eg.
```
   err = wait.PollImmediate(2*time.seconds, 2*time.minutes, func() (done bool, err error) {
             _, err = updateObjectBucket(ob, c.libClientset)
    	     if err != nil {
		     log.Error(err, "updating OB failed, retrying...")
		     return false, err
	     }
             return true, nil
    })

    if err != nil {
    	log.Error(err, "updating OB failed")
	// revert to old values
	overwriteAdditionalConfig(tmp, ob.Spec.Endpoint.AdditionalConfigData)
	err = c.provisioner.Update(ob)
    }
    return err
}
```

pkg/provisioner/controller.go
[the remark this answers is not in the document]
thotz: IMO `oldSpecWithNewAdditionalConfig` is pointer than a `Spec`. I have added following function to print `Spec` values
```diff
diff --git a/pkg/provisioner/controller.go b/pkg/provisioner/controller.go
index 68f47e9..ddc4cc7 100644
--- a/pkg/provisioner/controller.go
+++ b/pkg/provisioner/controller.go
@@ -105,6 +105,7 @@ func NewController(provisionerName string, provisioner api.Provisioner, clientse
                        oldSpecWithNewAdditionalConfig := oldObc.Spec.DeepCopy()
                        overwriteAdditionalConfig(newObc.Spec.AdditionalConfig, oldSpecWithNewAdditionalConfig.AdditionalConfig)
                        if !reflect.DeepEqual(newObc.Spec, oldObc.Spec) {
+                               printObcSpec(oldSpecWithNewAdditionalConfig)
                                // new OBC is changed
                                if !reflect.DeepEqual(*oldSpecWithNewAdditionalConfig, newObc.Spec) {
                                        // new OBC has changed something other than additionalConfig
@@ -695,3 +696,13 @@ func overwriteAdditionalConfig(srcAdditionalConfig map[string]string, destAdditi
                destAdditionalConfig[key] = value
        }
 }
+
+func printObcSpec(spec v1alpha1.ObjectBucketClaimSpec) {
+       logD.Info("storageclassName ", spec.StorageClassName)
+       logD.Info("bucketname ", spec.BucketName)
+       logD.Info("generatebucknam", spec.GenerateBucketName)
+       logD.Info("ObjectBucketName", spec.ObjectBucketName)
+       for key, value := range spec.AdditionalConfig {
+               logD.Info("key ", key, "value ", value)
+       }
+}
```
And give me following error :
```console
go build'ing package ./pkg/...
# github.com/kube-object-storage/lib-bucket-provisioner/pkg/provisioner
pkg/provisioner/controller.go:108:17: cannot use oldSpecWithNewAdditionalConfig (type *"github.com/kube-object-storage/lib-bucket-provisioner/pkg/apis/objectbucket.io/v1alpha1".ObjectBucketClaimSpec) as type "github.com/kube-object-storage/lib-bucket-provisioner/pkg/apis/objectbucket.io/v1alpha1".ObjectBucketClaimSpec in argument to printObcSpec
```

pkg/provisioner/controller.go
```diff
@@ -638,3 +648,62 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+	ob, err := getObjectBucketFromClaimKey(key, obc, c.libClientset, defaultRetryBaseInterval, defaultRetryTimeout)
+	if err != nil {
+		log.Error(err, "getting ob failed", "key", key)
+		return err
+	}
+
+	log.Info("Additional config changed, hence updating OB")
+	tmp := make(map[string]string)
+
+	tmp = ob.Spec.Endpoint.AdditionalConfigData
```
BlaineEXE, line 678: In Go, maps need to be deep copied. This will only change the pointer, and `tmp` will therefore be overwritten by `overwriteAdditionalConfig`. If updating the object bucket fails, this will not properly reset the value.

See my test here: https://play.golang.org/p/z0BurgSZEM8

Given this potential bug, it would be best to have a unit test that verifies the provisioner is properly reset when updating the OB fails.

BlaineEXE (changes requested):

pkg/provisioner/controller.go
```diff
@@ -638,3 +648,62 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
...
+			return false, err
+		}
+		return true, nil
+	})
+
+	if err != nil {
+		log.Error(err, "updating OB failed, reverting provisioner to original value")
+		overwriteAdditionalConfig(tmp, ob.Spec.Endpoint.AdditionalConfigData)
+		err = c.provisioner.Update(ob)
+	}
```
BlaineEXE, line 698: I think it's best to keep the `if err != nil` connected visually (no space between). I would separate the provisioner update from the OB update with a newline but not the error checking on updating the OB

```suggestion

	err = wait.PollImmediate(defaultRetryTimeout, defaultRetryBaseInterval, func() (done bool, err error) {
		_, err = updateObjectBucket(ob, c.libClientset)
		if err != nil {
			log.Error(err, "updating OB failed, retrying...")
			return false, err
		}
		return true, nil
	})
	if err != nil {
		log.Error(err, "updating OB failed, reverting provisioner to original value")
		overwriteAdditionalConfig(tmp, ob.Spec.Endpoint.AdditionalConfigData)
		err = c.provisioner.Update(ob)
	}
```

pkg/provisioner/controller.go
jeffvance: Yes you are right! Sorry @thotz

pkg/provisioner/controller.go
```diff
@@ -638,3 +648,62 @@ func (c *obcController) objectBucketForClaimKey(key string) (*v1alpha1.ObjectBuc
 	}
 	return ob, nil
 }
+
+func updateSupported(old, new *v1alpha1.ObjectBucketClaim) bool {
+	// The only field supported for update is obc.spec.additionalConfig
+	if reflect.DeepEqual(new.Spec, old.Spec) {
```
copejon, line 654: Is it necessary to compare the whole spec if the only supported field is additionalConfig?  Why not instead check if additionalConfig has changed?

jeffvance: One of the goals is to log an error or warning if the user attempts to change anything in the obc other than `.spec.additionalConfig`.  I guess DeepEqual handles map compares in cases where they keys may be in different order, right?

copejon: Nevermind, I believe it's just to short circuit the call if nothing call a non-spec field was changed.  Makes sense to me.

pkg/provisioner/controller.go
thotz: @BlaineEXE : I agree it is good to have a unit test. Sorry but I don't understand which function I need write the unit test.
Is it for `overwriteAdditionalConfig()` with set of string maps like in ur example. Or for the `handleUpdateClaim()`? 
If it is `handleUpdateClaim()`
First I need to create OBC and then call `handleUpdateClaim()` with that OBC for different test cases?

jeffvance (commented):
lgtm

(for future reference, I prefer to see pr changes in separate commits so that I can re-review only the code that has changed since my previous review)

thotz:
> lgtm
> 
> (for future reference, I prefer to see pr changes in separate commits so that I can re-review only the code that has changed since my previous review)

Sure I keep this in mind while sending PR. Sorry for the trouble and thanks for your patience while reviewing the PR

jeffvance:
No worries @thotz - single commit PRs are common, but, IMO, there are advantages to multi-commit PRs with sensible squashing just before merging.

pkg/provisioner/controller.go
BlaineEXE: I think for `handleUpdateClaim` and make sure the unit test coverage report activates the lines below
```go
if err != nil {
		log.Error(err, "updating OB failed, reverting provisioner to original value")
    		overwriteAdditionalConfig(tmp, ob.Spec.Endpoint.AdditionalConfigData)
    		err = c.provisioner.Update(ob)
}
```

BlaineEXE (approved):
I didn't realize I was the hold-up. Sorry. I think this looks good now.

leseb:
@thotz is Rook going to consume that?

thotz:
Yes I can update the library for Rook

PR #215, opened 2021-06-04
allow null OB subRegion in schema

### LINKED ISSUE #214: should honor empty `SubRegion`
`SubRegion` can be empty. It currently has the following definition
```yaml

type Endpoint struct {
        BucketHost           string            `json:"bucketHost"`
        BucketPort           int               `json:"bucketPort"`
        BucketName           string            `json:"bucketName"`
        Region               string            `json:"region"`
        SubRegion            string            `json:"subRegion"`
        AdditionalConfigData map[string]string `json:"additionalConfig"`
}```

That causes trouble when provisioner doesn't set it with meaningful value as in [here](https://github.com/rook/rook/blob/v1.6.4/pkg/operator/ceph/object/bucket/provisioner.go#L381)

It should cope with this situation by 

```yaml

type Endpoint struct {
        BucketHost           string            `json:"bucketHost"`
        BucketPort           int               `json:"bucketPort"`
        BucketName           string            `json:"bucketName"`
        Region               string            `json:"region"`
        SubRegion            string            `json:"subRegion,omitempty"`
        AdditionalConfigData map[string]string `json:"additionalConfig"`
}```

@leseb @copejon

copejon:
fixes #214 

Signed-off-by: jcope <jcope@redhat.com>

### DIFF
--- a/deploy/crds/objectbucket_v1alpha1_objectbucket_crd.yaml
+++ b/deploy/crds/objectbucket_v1alpha1_objectbucket_crd.yaml
@@ -96,6 +96,7 @@ spec:
                 subRegion:
                   description: Bucket sub-region
                   type: string
+                  nullable: true
                 additionalConfig:
                   description: AdditionalConfig gives providers a location to set
                     proprietary config values (tenant, namespace, etc)


leseb:
Out of curiosity why not set `,omitempty` to the type instead?

PR #216, opened 2021-08-13
fix OBC delete stuck

thotz:
The delete operation is handled via update calls, with addition of new
Update() API, the objects need to deleted are not added to the queue due
to check in updateSupported(). So if the deletion timestamp is set, then
this check is ignored in NewController().
FInd more details [here](https://github.com/rook/rook/pull/8514#issuecomment-898011184)


BlaineEXE (changes requested):
Why not make this part of `updateSupported()`?

BlaineEXE (approved):

PR #220, opened 2022-01-04
fixed bug in generateBucketName

dannyzaken:
when the prefix length is exactly maxBaseNameLen, the prefix is not truncated and as a result, the generated bucket name is invalid.

Signed-off-by: Danny Zaken <dannyzaken@gmail.com>

### DIFF
--- a/pkg/provisioner/helpers.go
+++ b/pkg/provisioner/helpers.go
@@ -197,7 +197,7 @@ const (
 )
 
 func generateBucketName(prefix string) string {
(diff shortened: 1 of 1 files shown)


BlaineEXE:
This looks good to me, but please clarify in the commit what exactly is making the generated name invalid. Is the generated name too long? Or is it an issue of not allowing multiple consecutive dashes? It's not fully clear to me.

BlaineEXE:
I did confirm that this patch fixes the issue in Rook upstream, and I do believe this is length-related.

PR #221, opened 2022-04-20
core: update go mod to get latest logr/klog

subhamkrai:
core: update go mod to get latest logr/klog

update go mod to get latest logr/klog.
These changes are required for Rook/Rook to consume
latest controller runtime version i.e v0.11.2

core: make changes to support latest klog/klogr
changes are required to support latest klog/klogr

Signed-off-by: subhamkrai <srai@redhat.com>


subhamkrai:
fyi, @BlaineEXE

BlaineEXE:
I think this looks good. I'm a bit concerned about why go.sum has 450 additions, but that would likely be a problem with some dependency I would think.

Issue #219, opened 2021-10-07
Panic in overwriteAdditionalConfig


romayalon:
Hey,
I used the latest changes of the update events of OBC, 
I noticed that if there is no additionalConfig in Spec before the update, and after the change there is, I get a panic error from `overwriteAdditionalConfig` function  in line `714` in `controller.go` and the pod is being restarted.
The error message is: `assignment to entry in nil map`, which sounds like an initialization issue. 

Thanks for your help

BlaineEXE:
```
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)                                                                                                                                    [33/1925]
        /go/pkg/mod/k8s.io/apimachinery@v0.22.2/pkg/util/runtime/runtime.go:48 +0x86
panic(0x1da4e60, 0x234d050)
        /usr/local/go/src/runtime/panic.go:965 +0x1b9
github.com/kube-object-storage/lib-bucket-provisioner/pkg/provisioner.overwriteAdditionalConfig(0xc00082a330, 0x0)
        /go/pkg/mod/github.com/kube-object-storage/lib-bucket-provisioner@v0.0.0-20220105185820-c1da9586e05b/pkg/provisioner/controller.go:714 +0xf8
github.com/kube-object-storage/lib-bucket-provisioner/pkg/provisioner.updateSupported(0xc000ba5500, 0xc000a3e480, 0x5)
        /go/pkg/mod/github.com/kube-object-storage/lib-bucket-provisioner@v0.0.0-20220105185820-c1da9586e05b/pkg/provisioner/controller.go:666 +0x133
github.com/kube-object-storage/lib-bucket-provisioner/pkg/provisioner.NewController.func1(0x1fda8e0, 0xc000ba5500, 0x1fda8e0, 0xc000a3e480)
        /go/pkg/mod/github.com/kube-object-storage/lib-bucket-provisioner@v0.0.0-20220105185820-c1da9586e05b/pkg/provisioner/controller.go:104 +0x8c
k8s.io/client-go/tools/cache.ResourceEventHandlerFuncs.OnUpdate(...)
        /go/pkg/mod/k8s.io/client-go@v0.22.2/tools/cache/controller.go:238
k8s.io/client-go/tools/cache.(*processorListener).run.func1()
        /go/pkg/mod/k8s.io/client-go@v0.22.2/tools/cache/shared_informer.go:775 +0x1c5
[...]

PR #222, opened 2022-08-09
fix nil pointer panic when changing bucket config

degorenko:
Related-Issue: https://github.com/kube-object-storage/lib-bucket-provisioner/issues/219

Signed-off-by: Denis Egorenko <degorenko@mirantis.com>

### DIFF
--- a/pkg/provisioner/controller.go
+++ b/pkg/provisioner/controller.go
@@ -663,7 +663,7 @@ func updateSupported(old, new *v1alpha1.ObjectBucketClaim) bool {
 	}
 	// create copy of old spec, and set the new spec's additionalConfig on it
 	oldspec := old.Spec.DeepCopy()
-	overwriteAdditionalConfig(new.Spec.AdditionalConfig, oldspec.AdditionalConfig)
+	oldspec.AdditionalConfig = new.Spec.AdditionalConfig
 	if !reflect.DeepEqual(*oldspec, new.Spec) {
 		// new OBC spec has changed something other than additionalConfig
 		log.Error(nil, "invalid changes to OBC. only additionalConfig can be updated")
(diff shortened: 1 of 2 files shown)


pkg/provisioner/controller.go
```diff
@@ -707,10 +707,14 @@ func (c *obcController) handleUpdateClaim(key string, obc *v1alpha1.ObjectBucket
...
-		destAdditionalConfig[key] = value
+	if destAdditionalConfig != nil {
+		for key, _ := range destAdditionalConfig {
+			delete(destAdditionalConfig, key)
+		}
+		for key, value := range srcAdditionalConfig {
+			destAdditionalConfig[key] = value
+		}
+	} else {
+		destAdditionalConfig = srcAdditionalConfig
```
BlaineEXE, line 718: I'm not sure if this is the source of the issue, but the copy must be a deep copy.

```suggestion
	if destAdditionalConfig != nil {
		for key, _ := range destAdditionalConfig {
			delete(destAdditionalConfig, key)
		}
	}
	for key, value := range srcAdditionalConfig {
		destAdditionalConfig[key] = value
	}
```

BlaineEXE (changes requested):

BlaineEXE:
I think this looks good. Would you describe the manual tests you used to test that the fix is working? I'd like to make sure there aren't any corner cases missed

degorenko:
@BlaineEXE  so, i did fork, made changes. But i'm currently using Rook 1.8.5, so i have to rebase on commit which is used in go.mod for 1.8.5 version. Finally, i have v0.0.0-20220105185820-c1da9586e05b plus that changes on top of it (I have additional config-override-1.8 branch in fork). Replaced in Rook go.mod l-b-p dependency on my fork commit, built Rook image on top of it and then test. So in my tests Rook 1.8.5+current change. Does it suites your expectations ?

BlaineEXE:
I was thinking to look at the logs and make sure...
- create OBC successful
- no changes to OBC (restart operator), nothing happens
- change OBC additional config, changes happen correctly now
- change another OBC field, controller errors saying it can't be updated
- delete OBC works

degorenko:
[rook-operator.log](https://github.com/kube-object-storage/lib-bucket-provisioner/files/9310460/rook-operator.log)
[user-stats.log](https://github.com/kube-object-storage/lib-bucket-provisioner/files/9310462/user-stats.log)

degorenko:
log contains some debug message around OBC spec diff, but it is not in current change, only for test purpose.

BlaineEXE:
The tests look great. Thank you. I'm trying to figure out why the CI is failing though.

BlaineEXE:
It looks like one of the tools used in CI is no longer hosted at the location. 

This line in travis.yml needs updated: https://github.com/kube-object-storage/lib-bucket-provisioner/blob/4bea5edaff58c0e80e3ad0c56f2894a7d456cc87/.travis.yml#L18 

Replacing that with this should get it going again I think.
```
curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/v1.16.0/install.sh | sh -s -- -b $(go env GOPATH)/bin && golangci-lint --version
```

BlaineEXE (approved):

BlaineEXE:
Merged. To pull the change into Rook, we just need a PR that updates the go.mod to use the latest version (hash) of this library. And we can have it backported back to 1.8 and 1.9 as well. 👍 Thanks again!

PR #224, opened 2022-10-31
BREAKING! allow OBs to be updated by provisioner

BlaineEXE:
THIS IS A BREAKING CHANGE!

Sometimes, an OB might need to be updated by the provisioner. For example, the endpoint of the object store changed.

In order for this to work, lib-bucket-provisioner must use an idempotent controller, and it must call Provision()/Grant() on each reconcile. Provision() and Grant() must then also be idempotent. In order to allow these calls to be more idempotent, l-b-p will now call GenerateUserID() to generate a deterministic idempotency key that Provision() and Grant() will then use as a user name and for any actions that need help being idempotent between runs.

Because Provision() and Grant() are idempotent, there is now no need for the Update() method.

This is a change to the previous API spec which allowed provisioners to return errors if a bucket already existed in the backend.

### DIFF
--- a/pkg/provisioner/api/provisioner.go
+++ b/pkg/provisioner/api/provisioner.go
@@ -26,6 +26,12 @@ import (
 // methods used to create and delete new buckets, and to grant or revoke access
 // to buckets within the object store.
 type Provisioner interface {
+	// GenerateUserID should deterministically generate a user ID for an OBC. This ID is used as an
+	// idempotency key in order to ensure repeat calls to Provision or Grant are consistent.
+	// Lib-bucket-provisioner may pass a non-nil ObjectBucket with the function as well. This will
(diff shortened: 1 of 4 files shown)


travisn:
The breaking change is that the API is not returning errors anymore where it would have previously? In that case, I would guess that clients such as Rook users would not see this as a breaking change under normal conditions?

BlaineEXE:
The breaking API change is that Provision() and Grant() are expected to be idempotent. That previously was not the case. As I'm working to implement this with Rook, I'm realizing there are other changes needed as well.

BlaineEXE:
To clarify some more about Provision() no longer requiring cleanup:

This library has generated the bucket name before calling Provision() for a while, and now with GetUserID() deterministically generating a user ID, there shouldn't be any risk of leaving dangling buckets/users with random names on the backend. For partially-provisioned OBCs, cleanup can now safely happen when Delete() is called.

PR #230, opened 2026-04-15
Fix json tags in ObjectBucketClaimSpec

karthikps97:
### Explain the changes:
1. Fix the typo in `omitempty` for `BucketName` field.
2. Add `omitempty` for `StorageClassName` field (we want to omit this field from the final generated json). We can also pass the json with an empty spec (not intended for OBC creation) with this change.

### DIFF
--- a/pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
+++ b/pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
@@ -31,13 +31,15 @@ func ObjectBucketClaimGVK() schema.GroupVersionKind {
 type ObjectBucketClaimSpec struct {
 
 	// StorageClass names the StorageClass object representing the desired provisioner and parameters
-	StorageClassName string `json:"storageClassName"`
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	StorageClassName string `json:"storageClassName,omitempty"`
 
 	// BucketName (not recommended) the name of the bucket.  Caution!
 	// In-store bucket names may collide across namespaces.  If you define
 	// the name yourself, try to make it as unique as possible.
 	// +optional
-	BucketName string `json:"bucketName,omityempty"`
+	BucketName string `json:"bucketName,omitempty"`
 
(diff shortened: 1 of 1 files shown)


BlaineEXE:
I think the typo fix is easily acceptable here, but I am concerned about your intention to create OBCs without specifying the required field: StorageClassName. What purpose do you have for omitting this?

pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
```diff
@@ -31,13 +31,13 @@ func ObjectBucketClaimGVK() schema.GroupVersionKind {
 type ObjectBucketClaimSpec struct {
 
 	// StorageClass names the StorageClass object representing the desired provisioner and parameters
-	StorageClassName string `json:"storageClassName"`
+	StorageClassName string `json:"storageClassName,omitempty"`
```
BlaineEXE, line 36: I have concerns about your usage when you say you want to generate JSON that won't include a required field. 

At any rate, this is a good improvement, and it can be made clearer by explicitly marking this field as required and also making sure the CRDs require minlength=1.

```suggestion
	// +required
	// +kubebuilder:validation:MinLength=1
	StorageClassName string `json:"storageClassName,omitempty"`
```

karthikps97:
@BlaineEXE, We have a scenario where we send the marshaled resource over a gRPC call and I want to omit the `spec` here.

pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
karthikps97: Yes, Adding the kubebuilder directive would allow us to implement the changes to json tag as well as making sure that `StorageClassName` is a required field.

BlaineEXE:
Running this through Rook's CI here: https://github.com/rook/rook/pull/17388

BlaineEXE:
Object test passed in Rook. Merging. Thanks @karthikps97

Issue #232, opened 2026-04-29
Additional data provided in OBC is not passed to config map related to bucket


degorenko:
API field `additionalConfig` provided in OBC is not passed to related config map. In some cases this extra data may contain additional env var.
As of now, need to manually check OB for stuff in that field and pass env vars to pods in different way from just using envFrom related config map.

BlaineEXE:
I'm not following. Could you be more specific?

degorenko:
Idea is pass in additionalConfig for OBC env vars, which then will be saved in config map, which is going to be used as env. Right now map contains only main vars like

```
			bucketName:      ep.BucketName,
			bucketHost:      ep.BucketHost,
			bucketPort:      strconv.Itoa(ep.BucketPort),
			bucketRegion:    ep.Region,
			bucketSubRegion: ep.SubRegion,
```

while can be required some other vars, based on IAM and app specific.

BlaineEXE:
Could you be more specific about what envs you are wanting to be added. Like what is the information that's missing that you want to add? Paint me an end-to-end picture please.

We can make changes to this repo, but it is in maintenance mode, so changes should be limited to bug fixes only. This seems like a feature addition to me, and I'd like to be able to assess whether this is something that makes sense to proceed forward with or not. And if not, I'd like to respond with a quality alternative suggestion.

degorenko:
So idea to pass env vars like AWS https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-envvars.html in configmap as single point of env vars source. Is it right place to pass such params? Or originally OBC additional config is not designed at all for such settings?

PR #233, opened 2026-04-29
Pass provided additional data config to config map

degorenko:
In some cases need to pass extra env vars which can be consumed by envFrom config map for bucket.

Related-Issue: #232

Signed-Off-By: Denis Egorenko <degorenko@mirantis.com>

### DIFF
--- a/pkg/provisioner/resourcehandlers.go
+++ b/pkg/provisioner/resourcehandlers.go
@@ -63,7 +63,7 @@ func newBucketConfigMap(obc *v1alpha1.ObjectBucketClaim, ep *v1alpha1.Endpoint, 
 		return nil, fmt.Errorf("cannot construct configMap, got nil OBC")
 	}
 
-	return &corev1.ConfigMap{
+	cm := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:       composeConfigMapName(obc),
 			Namespace:  obc.Namespace,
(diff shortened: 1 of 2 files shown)


BlaineEXE:
OBCs aren't really meant to be used in this manner. Could you be more specific about what you want to happen end to end, without making it abstract?

degorenko:
Idea is pass in additionalConfig for OBC env vars, which then will be saved in config map, which is going to be used as env. Right now map contains only main vars like

```
			bucketName:      ep.BucketName,
			bucketHost:      ep.BucketHost,
			bucketPort:      strconv.Itoa(ep.BucketPort),
			bucketRegion:    ep.Region,
			bucketSubRegion: ep.SubRegion,
```

while can be required some other vars, based on IAM and app specific.

> OBCs aren't really meant to be used in this manner

What is general purpose of that field, if that passed only to OB ?

BlaineEXE:
The purpose of the field is to pass additional information to the provisioner, i.e. Rook in your case. It serves the same purpose as `parameters` on a StorageClass, but able to be user-specified.

From the docs:

> additionalConfig gives providers a location to set proprietary config values (tenant, namespace...).

It isn't intended as a mechanism to pass through information from the OBC definition to the ConfigMap.

degorenko:
> The purpose of the field is to pass additional information to the provisioner, i.e. Rook in your case.

So it is not intended to use AWS env vars like https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-envvars.html at all?

BlaineEXE:
Correct

PR #234, opened 2026-08-26
obc: Add CEL validation rules for ObjectBucketClaimSpec

parth-gr:
Add validation rules to enforce that bucketName and generateBucketName are mutually exclusive and immutable once set.
  - Add CEL rule: bucketName is immutable once set
  - Add CEL rule: generateBucketName is immutable once set

### DIFF
--- a/pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
+++ b/pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
@@ -28,6 +28,10 @@ func ObjectBucketClaimGVK() schema.GroupVersionKind {
 }
 
 // ObjectBucketClaimSpec defines the desired state of ObjectBucketClaim
+// +kubebuilder:validation:XValidation:rule="!(has(self.bucketName) && self.bucketName != '' && has(self.generateBucketName) && self.generateBucketName != '')",message="bucketName and generateBucketName are mutually exclusive, only one can be set"
+// +kubebuilder:validation:XValidation:rule="(has(self.bucketName) && self.bucketName != '') || (has(self.generateBucketName) && self.generateBucketName != '')",message="one of bucketName or generateBucketName must be set"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.bucketName) || self.bucketName == oldSelf.bucketName",message="bucketName is immutable once set"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.generateBucketName) || self.generateBucketName == oldSelf.generateBucketName",message="generateBucketName is immutable once set"
 type ObjectBucketClaimSpec struct {
 
 	// StorageClass names the StorageClass object representing the desired provisioner and parameters


pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
BlaineEXE, line 52: Since these fields are both optional, I think these have to also use `has()`. I'm not 100% sure, but AFAIU, adding it won't hurt. You can test wether it works by seeing if the OBC's CRD allows removing `bucketName`/`generateBucketName` after it's specified.

There is a kube blog that does a good job of providing a known working pattern for immutability after being set: https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/#immutablility-after-first-modification

In short, we need to specify that both the value remains the same (on the field) and also that the field is present (on the field's parent struct).

(The git suggestion diff is rendering poorly, so readers will need to pay extra attention to actually see what fields I removed/added.)

```suggestion
// +kubebuilder:validation:XValidation:rule="!(has(self.bucketName) && self.bucketName != '' && has(self.generateBucketName) && self.generateBucketName != '')",message="bucketName and generateBucketName are mutually exclusive, only one can be set"
// +kubebuilder:validation:XValidation:rule="!has(oldSelf.bucketName) || has(self.bucketName)", message="bucketName is required once set"
// +kubebuilder:validation:XValidation:rule="!has(oldSelf.generateBucketName) || has(self.generateBucketName)", message="generateBucketName is required once set"
type ObjectBucketClaimSpec struct {

// StorageClass names the StorageClass object representing the desired provisioner and parameters
// +required
// +kubebuilder:validation:MinLength=1
StorageClassName string `json:"storageClassName,omitempty"`

// BucketName (not recommended) the name of the bucket.  Caution!
// In-store bucket names may collide across namespaces.  If you define
// the name yourself, try to make it as unique as possible.
// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="bucketName is immutable"
// +optional
BucketName string `json:"bucketName,omitempty"`

// GenerateBucketName (recommended) a prefix for a bucket name to be
// followed by a hyphen and 5 random characters. Protects against
// in-store name collisions.
// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="generateBucketName is immutable"
// +optional
GenerateBucketName string `json:"generateBucketName,omitempty"`

```

Aside: IMO, it would also be reasonable to add a rule that requires at least one of the fields to be set, but this could possibly be a breaking change. I'm not sure if we could add it safely here.

BlaineEXE (changes requested):

pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
parth-gr: Is it possible to create a bucket if no field is set?

parth-gr: It wont be a breaking change
As either one of this required,
```
E0903 07:59:07.741561       1 controller.go:204] "Unhandled Error" err="error syncing 'rook-ceph/ceph-retain-bucket1': error composing bucket name: expected either bucketName or generateBucketName defined, requeuing" logger="UnhandledError"

```

parth-gr: I added once set it cant be change, but once we add the rule atleast one can be set, the check for has doesnt add up anything, as if cr is created one should be always set

deploy/crds/v1/objectbucket.io_objectbucketclaims.yaml
```diff
@@ -0,0 +1,113 @@
+---
+apiVersion: apiextensions.k8s.io/v1
```
BlaineEXE: Why is this upgrading the CRDs to v1? They are still v1alpha1 and should update the CRDs already present in `in lib-bucket-provisioner/deploy
/crds/`

BlaineEXE (changes requested):

hack/update-crds.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+# Generates the v1 CRD manifests from the API types in
```
BlaineEXE: OBCs aren't v1

hack/update-crds.sh
```diff
@@ -0,0 +1,12 @@
...
+# pkg/apis/objectbucket.io/v1alpha1/*types.go using controller-gen.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+controller-gen crd \
+    paths="./pkg/apis/objectbucket.io/v1alpha1/..." \
+    output:crd:dir=deploy/crds/v1
```
BlaineEXE: We may also need to modify the args to match the filenames already in use

```suggestion
controller-gen crd \
    paths="./pkg/apis/objectbucket.io/v1alpha1/..." \
    output:crd:dir=deploy/crds
```

BlaineEXE (changes requested):

deploy/crds/v1/objectbucket.io_objectbucketclaims.yaml
parth-gr: In Rook, we are already using https://github.com/rook/rook/blob/master/deploy/examples/crds.yaml#L17551 v1,
And The CEL rules are not present in v1alpha1.

hack/update-crds.sh
parth-gr: Creating a v1 was necessary, 
THis is the standard filename generated by controller gen for v1, 
But we can customise it if required,
So, creating a v1 separate folder is fine? And then maybe I can pick the naming convention of v1alpha1 file

hack/update-crds.sh
parth-gr: Not as per rook codebase https://github.com/rook/rook/blob/master/deploy/examples/crds.yaml#L17551

deploy/crds/v1/objectbucket.io_objectbucketclaims.yaml
```diff
@@ -0,0 +1,113 @@
...
+      name: StorageClass
+      type: string
+    - description: Phase
+      jsonPath: .status.phase
+      name: Phase
+      type: string
+    - jsonPath: .metadata.creationTimestamp
+      name: Age
+      type: date
+    name: v1alpha1
```
parth-gr: @BlaineEXE here the version is v1alpha1, thats the api version change at the top, for using cel rules

deploy/crds/v1/objectbucket.io_objectbucketclaims.yaml
BlaineEXE: Oops. My bad. This is the apiextensions API, not OB/OBC.

However, we still should update the CRDs already present that are in `deploy/crd`. Putting these in `/v1` subdir (I think?) suggests that these are v1 of the OB/OBC library.

hack/update-crds.sh
BlaineEXE: Let's just update the CRDs already in place. There's no need to keep them

parth-gr: Updated, it  inplace

deploy/crds/objectbucket_v1alpha1_objectbucketclaim_crd.yaml
```diff
@@ -1,88 +1,113 @@
...
-          description: Most recently observed status of the claim.
-          properties:
-            phase:
-              description: ObjectBucketClaimStatusPhase is set by the controller to save the state of the provisioning process
-              enum:
-                - "Pending"
-                - "Bound"
-                - "Released"
-                - "Failed"
-              type: string
```
BlaineEXE, line 87: It looks like the enum present in the original isn't making it to the new generated CRD. Let's add these to the types as needed so the generation keeps the old content as closely as possible

deploy/crds/objectbucket_v1alpha1_objectbucketclaim_crd.yaml
```diff
@@ -1,88 +1,113 @@
...
-      - obc
-      - obcs
+    - obc
+    - obcs
+    singular: objectbucketclaim
   scope: Namespaced
-  subresources:
-    status: {}
-  additionalPrinterColumns:
-  - JSONPath: .spec.storageClassName
```
BlaineEXE, line 24: This old CRD is a mess. I wonder if someone hand wrote parts of these.

pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go
BlaineEXE, line 84: Needs enum definition to keep new CRD from diverging from old one. Ditto for ObjectBucketStatus in the other file.

deploy/crds/objectbucket_v1alpha1_objectbucket_crd.yaml
```diff
@@ -1,126 +1,172 @@
...
-              description: StorageClass names the StorageClass object representing the 
-                desired provisioner and parameters
-              type: string
-            reclaimPolicy:
-              description: Describes a policy for end-of-life maintenance of ObjectBucket.
-              enum:
-                - "Delete"
-                - "Retain"
-                - "Recycle"
-              type: string
```
BlaineEXE, line 76: I think this needs an enum def as well

parth-gr: its a `  *corev1.PersistentVolumeReclaimPolicy` we cant update the k8s api IMO

parth-gr: but adding in the spec

BlaineEXE: Thanks for finding a way to keep those. Keeping as much of the existing CRD spec in place should help us avoid conflicts with Noobaa as well.

BlaineEXE (approved):
This looks good. Thanks for the detailed attention. Let's make sure the Rook CI passes with these changes before merging. That's my usual process for l-b-p updates that Rook consumes

pkg/provisioner/controller.go
```diff
@@ -449,7 +449,7 @@ func (c *obcController) handleDeleteClaim(key string, obc *v1alpha1.ObjectBucket
 
 	// call Delete or Revoke and then delete generated k8s resources
 	// Note: if Delete or Revoke return err then we do not try to delete resources
-	ob, err := updateObjectBucketPhase(c.libClientset, ob, v1alpha1.ObjectBucketClaimStatusPhaseReleased)
+	ob, err := updateObjectBucketPhase(c.libClientset, ob, v1alpha1.ObjectBucketStatusPhaseReleased)
```
parth-gr, line 452: @BlaineEXE as there is change in the file, we need to get this merged and then update the rook go mod to use it

deploy/crds/objectbucket_v1alpha1_objectbucketclaim_crd.yaml
```diff
@@ -1,88 +1,119 @@
...
-                - "Released"
-                - "Failed"
-              type: string
-          type: object
+            type: object
+            x-kubernetes-validations:
+            - message: bucketName and generateBucketName are mutually exclusive, only
+                one can be set
+              rule: '!(has(self.bucketName) && self.bucketName != '''' && has(self.generateBucketName)
+                && self.generateBucketName != '''')'
```
BlaineEXE, line 93: If we remove this rule, do we stop having issues with the OBC reconciler? 

It seems like the biggest change we want/need is to make sure the fields are immutable once set. IIRC, this rule wasn't enforced before, so maybe we need to relax it back.

parth-gr:
Testing done:
```
rider:examples$ oc apply -f object-bucket-claim-a.yaml 
Warning: resource objectbucketclaims/ceph-bucket-a2 is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by oc apply. oc apply should only be used on resources created declaratively by either oc create --save-config or oc apply. The missing annotation will be patched automatically.
The ObjectBucketClaim "ceph-bucket-a2" is invalid: spec: Invalid value: bucketName is immutable once set

```

deploy/crds/objectbucket_v1alpha1_objectbucket_crd.yaml
```diff
@@ -1,126 +1,181 @@
...
+                    type: string
+                  subRegion:
+                    type: string
+                required:
+                - additionalConfig
+                - bucketHost
+                - bucketName
+                - bucketPort
+                - region
+                - subRegion
```
BlaineEXE: It looks like these are newly being added as required. I am concerned that changing fields from optional to required might break existing OBs/OBCs, especially Noobaa ones that we are less experienced with. 

We may need to add `// +optional` tags to a bunch of fields on OBs. Or we might set optional as the default behavior and add `// +required` tags to the required fields instead. Either way would work; it would be best to follow whatever OBC types define, which seem to be generating properly now

deploy/crds/objectbucket_v1alpha1_objectbucket_crd.yaml
```diff
@@ -1,126 +1,181 @@
...
-              type: object
-          required:
+              storageClassName:
+                type: string
+            required:
+            - additionalState
+            - claimRef
+            - endpoint
+            - reclaimPolicy
             - storageClassName
```
BlaineEXE, line 150: Looks like `storageClassName` was the only previously-required field here

deploy/crds/objectbucket_v1alpha1_objectbucketclaim_crd.yaml
```diff
@@ -1,88 +1,112 @@
...
+                  ObjectBucketName is the name of the object bucket resource. This is the authoritative
+                  determination for binding.
+                type: string
+              storageClassName:
+                description: StorageClass names the StorageClass object representing
+                  the desired provisioner and parameters
+                minLength: 1
+                type: string
+            required:
             - storageClassName
```
BlaineEXE, line 87: Looks like the required fields on OBCs are generating well. No need to change this.

deploy/crds/objectbucket_v1alpha1_objectbucket_crd.yaml
```diff
@@ -1,126 +1,181 @@
...
+              phase:
+                description: ObjectBucketStatusPhase is set by the controller to save
+                  the state of the provisioning process.
+                enum:
+                - Bound
+                - Released
+                - Failed
+                type: string
+            required:
+            - phase
```
BlaineEXE: Looks like this is newly required

BlaineEXE (changes requested):

deploy/crds/objectbucket_v1alpha1_objectbucket_crd.yaml
parth-gr: Thanks for catching, updated it

BlaineEXE (approved):
Great! Thank you for the detailed attention here and for writing the script to help us keep this up-to-date better than we have in the past.

PR #235, opened 2026-09-04
ObjectBucket CRD schema missing claimRef fields

naveenpaul1:
### Describe the Problem
ObjectBucket CRD schema missing `claimRef` fields leads to data loss and bucket hijacking

### Explain the changes
1. add 'x-kubernetes-preserve-unknown-fields" OB CRD

### Issues: Fixed #xxx / Gap #xxx
1. https://redhat.atlassian.net/browse/DFBUGS-7972

### Testing Instructions:
1. create OB and OBC and verify `claimRef` added with valid object

- [ ] Doc added/updated
- [ ] Tests added


deploy/crds/objectbucket_v1alpha1_objectbucket_crd.yaml
```diff
@@ -77,6 +77,8 @@ spec:
             claimRef:
               description: ObjectReference to ObjectBucketClaim
               type: object
+              nullable: true
```
parth-gr, line 80: We also need to update the types.go,

And with https://github.com/kube-object-storage/lib-bucket-provisioner/pull/234 we are updating the crd api

naveenpaul1: @parth-gr will `types.go` too

PR #236, opened 2026-09-17
obc: give some relaxation to create obc with empty specs

parth-gr:
this change is for noobaa for allowing to update specs with empty string speically the bucket name


BlaineEXE (approved):

BlaineEXE:
Thanks @parth-gr . Let's also copy over the change to Rook, but I don't think this small change is particularly urgent.

